PR #2, opened 2021-03-06
Add hourglass architecture and some structure to the repo

robertklee:
Add a hourglass architecture with a variable number of stacks, variable input size, etc.

### DIFF
--- a/hourglass.py
+++ b/hourglass.py
@@ -31,3 +31,15 @@ class HourglassNet(object):
         # show model summary and layer name
         if show:
             self.model.summary()
+        
+    def train(self, batch_size, model_path, epochs):
+        # TODO
+        pass
+    
+    def resume_train(self, batch_size, model_json, model_weights, init_epoch, epochs):
+        # TODO
+        pass
+
+    def load_model(self, modeljson, modelfile):
+        # TODO
+        pass
--- a/hourglass.py
+++ b/hourglass.py
@@ -1,0 +1,30 @@
+import os
+from hourglass_blocks import create_hourglass_network, bottleneck_block, bottleneck_mobile
+from keras.callbacks import CSVLogger, ModelCheckpoint
+from keras.models import load_model, model_from_json
+from keras.optimizers import Adam, RMSprop
+from keras.losses import mean_squared_error
+import datetime
+import scipy.misc
+import numpy as np
+
+
+class HourglassNet(object):
+
+    def __init__(self, num_classes, num_stacks, num_channels, inres, outres):
+        self.num_classes = num_classes
+        self.num_stacks = num_stacks
+        self.num_channels = num_channels
+        self.inres = inres
+        self.outres = outres
+
+    def build_model(self, mobile=False, show=False):
+        if mobile:
+            self.model = create_hourglass_network(self.num_classes, self.num_stacks,
+                                                  self.num_channels, self.inres, self.outres, bottleneck_mobile)
+        else:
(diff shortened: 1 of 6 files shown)


hourglass.py
```diff
@@ -0,0 +1,45 @@
...
+                                                  self.num_channels, self.inres, self.outres, bottleneck_mobile)
+        else:
+            self.model = create_hourglass_network(self.num_classes, self.num_stacks,
+                                                  self.num_channels, self.inres, self.outres, bottleneck_block)
+        # show model summary and layer name
+        if show:
+            self.model.summary()
+        
+    def train(self, batch_size, model_path, epochs):
+        # TODO
```
julianrocha, line 36: Assuming this is where we would interface with the input pipeline. Something like:
```python
model = # load model from model_path and compile etc...
train_gen = DataGenerator("cleaned_dataframe.csv", "data", (244, 244), batch_size=batch_size, shuffle=True)
val_gen = DataGenerator(...)
model.fit(train_gen,val_gen,epochs=epochs...)
```
[Reference](https://mahmoudyusof.github.io/facial-keypoint-detection/data-generator/)

hourglass_blocks.py
```diff
@@ -0,0 +1,193 @@
+import keras.backend as K
+from keras.layers import *
+from keras.losses import mean_squared_error
+from keras.models import *
+from keras.optimizers import Adam, RMSprop
+
+# Adapted from https://github.com/yuanyuanli85/Stacked_Hourglass_Network_Keras/blob/master/src/net/hg_blocks.py
+
+def create_hourglass_network(num_classes, num_stacks, num_channels, inres, outres, bottleneck):
+    input = Input(shape=(inres[0], inres[1], 3))
```
julianrocha, line 10: Configurable input size, 3 channels, any idea what dimensions would work best for this arch?

hourglass_blocks.py
```diff
@@ -0,0 +1,193 @@
...
+                  name=str(hgid) + '_conv_1x1_x2')(head)
+    head_m = Conv2D(num_channels, kernel_size=(1, 1), activation='linear', padding='same',
+                    name=str(hgid) + '_conv_1x1_x3')(head_parts)
+
+    head_next_stage = Add()([head, head_m, prelayerfeatures])
+    return head_next_stage, head_parts
+
+
+def euclidean_loss(x, y):
+    return K.sqrt(K.sum(K.square(x - y)))
```
julianrocha, line 193: Is this a temporary loss function, will we be replacing this?

scripts/coco_dl.sh
```diff
@@ -0,0 +1,38 @@
+# Credit to https://gist.github.com/mkocabas/a6177fc00315403d31572e17700d7fd9
```
julianrocha, line 1: nice

julianrocha (commented):
Great work Robert this looks awesome. Some of the details in `hourglass_blocks.py` are a little confusing to me so wouldn't mind a brief overview in the meeting tn.

hourglass_blocks.py
robertklee: yep, these are given in `train.py`. Num classes is the number of joints, so for us it is 17. Stacks should be 2-8, with the paper we are basing it off of using 8.

hourglass_blocks.py
robertklee: Temporary for now

hourglass.py
robertklee: Yep, there's still a bit to be written here. Mostly data generator, callbacks, model saving.

PR #3, opened 2021-03-07
Input Pipeline Mock 1

julianrocha:
This PR addresses the issue of loading the COCO dataset for the model. 

DataGenerator:
Module to load images/annotations in batches. Currently handles scaling RGB channels, cropping to bbox, resizing images fixed dimensions, and generating heatmaps. We will eventually need to add online image augmentation, etc.

notebooks/generate_annCSV:
Notebook to generate a CSV of COCO annotations. This is a subset of COCO (~15'000 images) which contain only one person and at least 10 key points. This csv is needed for the data generator.

notebooks/input_pipeline_demo.ipynb
This notebook demos how to use the data generator and demos how to create symlinks in a collab runtime as well as how to prepopulate the filesystem cache with the COCO directories.


data_generator.py
```diff
@@ -0,0 +1,76 @@
...
+    return cv2.resize(img,self.output_size,interpolation=cv2.INTER_LINEAR) # what interpolation is best?
+
+  # adjust x/y coords to new resized img
+  def resize_label(self,label,og_width,og_height):
+    x_scale = self.output_size[0]/og_width
+    y_scale = self.output_size[1]/og_height
+    return [int(round(int(val)*(x_scale if i%3==0 else y_scale if i%3==1 else 1))) for i,val in enumerate(label)]
+
+  # returns batch at index idx
+  def __getitem__(self, idx):
```
robertklee, line 122: have you considered making this a generator function (using `yield`)?

robertklee: Never mind, using `Sequence` means you don't need `yield` per [stack overflow](https://stackoverflow.com/questions/56079223/custom-keras-data-generator-with-yield)

data_generator.py
```diff
@@ -0,0 +1,76 @@
...
+  def __len__(self):
+    return int(len(self.df) / self.batch_size)
+
+  # scale RGB channels to [0,1]
+  def scale_input(self,img):
+    return img/255.0 
+
+  # resize img to input shape (will likely replace once we start dealing with bboxes)
+  def resize_img(self,img):
+    return cv2.resize(img,self.output_size,interpolation=cv2.INTER_LINEAR) # what interpolation is best?
```
robertklee: Stack Overflow suggests `INTER_LINEAR` is a good place to start: https://stackoverflow.com/questions/23853632/which-kind-of-interpolation-best-for-resizing-image

data_generator.py
```diff
@@ -0,0 +1,76 @@
...
+
+  # adjust x/y coords to new resized img
+  def resize_label(self,label,og_width,og_height):
+    x_scale = self.output_size[0]/og_width
+    y_scale = self.output_size[1]/og_height
+    return [int(round(int(val)*(x_scale if i%3==0 else y_scale if i%3==1 else 1))) for i,val in enumerate(label)]
```
robertklee: what is this line doing?

data_generator.py
```diff
@@ -0,0 +1,76 @@
...
+
+  # returns batch at index idx
+  def __getitem__(self, idx):
+    ## Initialize Batch:
+    X = np.empty((self.batch_size, *self.output_size, 3)) # 3 channels of RGB
+    y = np.empty((self.batch_size, 3 * 17)) # (x,y,v) * 17 keypoints
```
robertklee: can we move the 3 and 17 to our `constants.py` class? (There's already an entry for num_keypoints)

data_generator.py
```diff
@@ -0,0 +1,76 @@
...
+import cv2
+import matplotlib.image as mpimg
+import os
+import tensorflow as tf
+import skimage.io as io
+from tensorflow.keras.utils import Sequence
+
+class DataGenerator(Sequence):
+
+  def __init__(self, csv_file, base_dir, output_size, shuffle=False, batch_size=10):
```
robertklee: need to pass in `num_hg_blocks` since our labels need to be duplicated based on the number of hourglass blocks (and thus the number of intermediate supervision loss calculations we need to do)

data_generator.py
robertklee: (sorry just hard to unpack what's going on)

data_generator.py
robertklee: y should be `(self.batch_size, constants.output_dim <you might need to make this variable, self.num_hg_blocks * constants.num_keypoints)` in dimension I think

data_generator.py
```diff
@@ -0,0 +1,76 @@
...
+    X = np.empty((self.batch_size, *self.output_size, 3)) # 3 channels of RGB
+    y = np.empty((self.batch_size, 3 * 17)) # (x,y,v) * 17 keypoints
+
+    # get the indices of the requested batch
+    indices = self.indices[idx*self.batch_size:(idx+1)*self.batch_size]
+
+    for i, data_index in enumerate(indices):
+      # Load label and img
+      ann = self.df.loc[data_index]
+      img = mpimg.imread(os.path.join(self.base_dir,ann['path'])) # bottleneck
```
robertklee: I'm wondering if we should use opencv for image read:

[source](https://towardsdatascience.com/what-library-can-load-image-in-python-and-what-are-their-difference-d1628c6623ad)

```
+------------+--------------+---------+
|  Library   |   Function   |  Time   |
+------------+--------------+---------+
| matplotlib | plt.imread() | 0.02254 |
| OpenCV     | cv2.imread() | 0.01096 |
| Pillow     | Image.Open() | 0.00166 |
| Skimage    | io.imread()  | 0.01463 |
+------------+--------------+---------+
```

robertklee: `matplotlib` seems to be the slowest of the four options. I'm not sure if Pillow is friendly with `numpy` since I've never used it before for this purpose.

data_generator.py
```diff
@@ -0,0 +1,76 @@
...
+
+      # Preprocess img
+      img = DataGenerator.scale_input(self,img)
```
robertklee: just for ease of future additions, maybe we could do a `self.transform_image` and `self.transform_labels`? 

Also, any reason why we don't use `self.scale_input` `self.resize_img` etc?

robertklee: we can pass in any scale, rotate, translation, brightness, etc

robertklee (commented):
I'm going to add in a feature to auto-convert python notebooks to .py for code review simplicity

robertklee:
[Here](https://app.reviewnb.com/robertklee/COCO-Human-Pose/pulls/) is a resource to review Python notebooks!

notebooks/generate_annCSV.ipynb
```diff
@@ -0,0 +1 @@
...convert_to_df(train_coco, 'train2017')     \n","train_coco_df = pd.merge(images_df, persons_df, right_index=True, left_index=True)\n","train_coco_df['source'] = 0\n","train_coco_df.head()\n","\n","images_df, persons_df = convert_to_df(val_coco, 'val2017')      \n","val_coco_df = pd.merge(images_df, persons_df, right_index=True, left_index=True)\n","val_coco_df['source'] = 1\n","val_coco_df.head()\n","\n","coco_df = pd.concat([train_coco_df, val_coco_df], ignore_index=True)\n","# ^ Dataframe containing all val and test keypoint annotations\n","\n","#cleaned_coco_df\n","cleaned_coco_df = coco_df.loc[coco_df['num_keypoints'] >3].reset_index()\n","cleaned_coco_df"],"execution_count":null,"outputs":[]},{"cell_type":"code","metadata":{"id":"xtUj7_dUuozM"},"source":["anns_per_img = coco_df.groupby('src_set_image_id',as_index=False).size()\n","imgs_w_one_ann = anns_per_img.loc[anns_per_img['size'] == 1]\n","imgs_w_one_ann = pd.merge(imgs_w_one_ann,coco_df)\n","one_ann_min10kp_imgs = imgs_w_one_ann.loc[imgs_w_one_ann['num_keypoints'] > 10]\n","one_ann_min10kp_imgs\n","csv_data = one_ann_min10kp_imgs.to_csv('data/one_ann_min10kp_imgs.csv',index=False)"],"execution_count":null,"outputs":[]}]}
```
robertklee: <p>I think if possible we could move this code to a python file, and then call that file from the colab notebook?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/robertklee/COCO-Human-Pose/pull/3/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='0'/>

notebooks/input_pipeline_demo.ipynb
```diff
@@ -0,0 +1 @@
...tories for fast and predictable load times\n","import os\n","def print_files_in_dir(path):\n","  print(len(os.listdir(path)), \" files at \", path)\n","\n","for i in range(0,6):\n","  print_files_in_dir('/content/gdrive/MyDrive/COCO/2017/val2017_sub-1000/' + str(i))\n","\n","for i in range(0,119):\n","    print_files_in_dir('/content/gdrive/MyDrive/COCO/2017/train2017_sub-1000/' + str(i))"],"execution_count":null,"outputs":[]},{"cell_type":"code","metadata":{"id":"8eqsEyXetzJV","executionInfo":{"status":"ok","timestamp":1615144372457,"user_tz":480,"elapsed":664,"user":{"displayName":"Julian Rocha","photoUrl":"https://lh3.googleusercontent.com/a-/AOh14Gi0QrES-WzNCHbIVd-DJKGaCtIC3OO5amuVSCFOPw=s64","userId":"17506259248702060718"}}},"source":["from data_generator import DataGenerator\n","\n","# Example Usage of Data Generator\n","train_gen = DataGenerator(\"data/one_ann_min10kp_imgs.csv\",\n","                          \"/content/datasets/coco\",\n","                          (256, 256),\n","                          batch_size=1,\n","                          shuffle=True)\n","# now let's train the model\n","#model.fit(train_gen, epochs=5, ...)"],"execution_count":91,"outputs":[]}]}
```
robertklee: <p>this is great! comments are 10/10</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/robertklee/COCO-Human-Pose/pull/3/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='1'/>

data_generator.py
julianrocha: This will be changed once we do proper cropping to bbox rather than stretching the whole image to input dimensions. This line is taking the label which looks like [x1,y1,v1,x2,y2,v2...] and fixing the x/y coords to match the new stretched image.

data_generator.py
julianrocha: Ok good for now then. This won't be necessary once we crop properly instead of stretching.

data_generator.py
julianrocha: @robertklee 
So to clarify, the desired output shape of the data generator is:

train_gen[0] --> X_batch, y_batch --> (batch_size,256,256,3), (batch_size,num_hg_blocks,3*17)

julianrocha: Is this duplication of data necessary for the model? Any reason why it can't reuse labels?

data_generator.py
julianrocha: I will move the magic numbers to constants.py but see other comment thread for shape questions.

data_generator.py
julianrocha: Great find, I will try the others out.

data_generator.py
julianrocha: Ya that's a cleaner way for sure haha, it's been a while since I've worked with Python classes so I was feeling pretty self.dumb ;)

notebooks/generate_annCSV.ipynb
julianrocha: <p>Yup good call. I'll make a module that can grab a df of all coco annotations. Then in a notebook we'll be able to filter for the anns we want before writing to csv.</p>

data_generator.py
robertklee: We'll still need it for crop and resize actually :)

data_generator.py
robertklee: > Is this duplication of data necessary for the model? Any reason why it can't reuse labels?

necessary because keras doesn't know they're the same labelling. I think it's just easier for keras to handle if we duplicate the ground truth labelling. they probably also do some fancy multi-core stuff with it.

robertklee: > train_gen[0] --> X_batch, y_batch --> (batch_size,256,256,3), (batch_size,**num_hg_blocks**, **num_hg_blocks** *17)
See inline changes

robertklee: X is right, y should be:

`(self.batch_size, constants.output_dim, self.num_hg_blocks * constants.num_keypoints)`

Output dim is the `64x64` heat map our model outputs (stored in `constants.py`) for each joint

We generate a heat map for *each* joint. so for a single ground truth evaluation, we have `64x64x17`.

Since we do intermediate supervision, we duplicate that n times, where n=num hg blocks. So this turns each image label to `64x64x(17*num_hg_blocks)`.

Then we do that for each image in the batch to get the final size of `batch_size x *output_dimension x (17*num_hg_blocks)`


Edit for clarity

robertklee: This will be used by keras for MSE evaluation.

notebooks/generate_annCSV.ipynb
robertklee: yep. Ideally i think our `train.py` and `eval.py` should handle everything end-to-end so the end user needs to just clone, `python3 train.py <arguments>`, and then `python3 eval.py <arguments>`. 

IMO, let's keep this in mind as we develop, and try to make things modular enough that we can implement this later on.

robertklee: Of course our Google Colab filesystem issues will make this not work lol

data_generator.py
julianrocha: Thanks for the clarification. Data duplication for keras makes sense then. Looks like the pipeline won't be usable till we get a heatmap label generator then.

robertklee: I can tackle that tomorrow

data_generator.py
```diff
@@ -33,44 +35,37 @@ def on_epoch_end(self):
...
+    transformed_label = [int(round(int(val)*(x_scale if i%NUM_COCO_KP_ATTRBS==0 else y_scale if i%NUM_COCO_KP_ATTRBS==1 else 1))) for i,val in enumerate(label)]
+    return np.asarray(transformed_label)
 
   # returns batch at index idx
   def __getitem__(self, idx):
-    ## Initialize Batch:
+    # Initialize Batch:
     X = np.empty((self.batch_size, *self.output_size, 3)) # 3 channels of RGB
-    y = np.empty((self.batch_size, 3 * 17)) # (x,y,v) * 17 keypoints
+    y = np.empty((self.batch_size, NUM_COCO_KP_ATTRBS * NUM_COCO_KEYPOINTS))
```
robertklee:
```suggestion
    y = np.empty((self.batch_size, *OUTPUT_DIM, NUM_COCO_KP_ATTRBS * NUM_COCO_KEYPOINTS))
```

robertklee: keep in mind for each frame we have a heat map of size `OUTPUT_DIM`

julianrocha: This change should fold in when heatmap generation gets added, otherwise, pipeline breaks b/c label shape doesn't match yet.

data_generator.py
julianrocha: Heatmap gen has been added, output shape adjusted to:
`y.shape = (self.batch_size, *self.output_dim, self.num_hg_blocks * constants.num_keypoints)`

data_generator.py
julianrocha: Change of shape added, resolving conversation

julianrocha:
<p>Thanks for the comments on this. These are actually old questions, from back when I was doing exploratory analysis of coco. That being said, the current bounding box cropping features of the data generator should easily be able to handle multi-person. Also, the CSV rows are unique on annotation ID, so image paths are already duplicated. No need to dynamically change the size of batch, because the batch size denotes the number of coco-annotations, not the number of images. This allows us to hold the property of constant batch size at the detriment of loading an image for each annotation it contains. </p><p> </p><p> </p><p>Zeroed heatmaps are being generated for unlabelled kps. Now feeding the model images that do not contain people (0 kps), we can likely load non category=1 images from coco. This feature could be added by creating a dummy annotation in the CSV of all 0's for (x,y,v)*17</p>

---

 <a href='https://app.reviewnb.com/robertklee/COCO-Human-Pose/pull/3/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

robertklee:
This is great! Some small suggestions for future iterations:
- make heatmap at full resolution and then downscale to reduce issue of heat map centering 
- bounding box + x percent size (maybe 130% of bounding box size), and then varied through data augmentation 110% to 150%?

nicolepeverley:
Hey @julianrocha  this looks great to me! I did some testing and found what parameters the coco evaluation api needs in order to run the object keypoint similarity precision metrics and put them into a json. I think they should be fairly easy to add as meta_data for the output, I'm assuming the bbox can be the bounding box you created and the keypoints are back in the regular scale. Let me know if you have any questions about it.

Here's the json format for the data/annotations: https://drive.google.com/file/d/1lodzJHapvJIanBoK_WbU9CjKa-gXOPSE/view?usp=sharing

- Adding the coco_url parameter is just so that we can put the keypoints on the original image for a visual representation
- The categories section can just be added as is at the end

robertklee:
@nicolepeverley  @julianrocha Prettified the JSON:

```
{
  "images": [
    {
      "coco_url": "http://images.cocodataset.org/val2014/COCO_val2014_000000001292.jpg",
      "id": 1292
    }
  ],
  "annotations": [
    {
      "num_keypoints": 15,
      "area": 75630.68985,
      "iscrowd": 0,
      "keypoints": [
        296,
        105,
        2,
        299,
        92,
        2,
        282,
        109,
        2,
        338,
        81,
        2,
        0,
        0,
        0,
        388,
        142,
        2,
        284,
        176,
        2,
        425,
        220,
        2,
        254,
        277,
        2,
        312,
        164,
        2,
        257,
        180,
        2,
        407,
        363,
        2,
        314,
        365,
        2,
        410,
        308,
        2,
        234,
        386,
        2,
        207,
        330,
        2,
        0,
        0,
        0
      ],
      "image_id": 1292,
      "bbox": [107.23, 47.68, 359.98, 379.32],
      "category_id": 1,
      "id": 1
    }
  ],
  "categories": [
    {
      "supercategory": "person",
      "name": "person",
      "skeleton": [
        [16, 14],
        [14, 12],
        [17, 15],
        [15, 13],
        [12, 13],
        [6, 12],
        [7, 13],
        [6, 7],
        [6, 8],
        [7, 9],
        [8, 10],
        [9, 11],
        [2, 3],
        [1, 2],
        [1, 3],
        [2, 4],
        [3, 5],
        [4, 6],
        [5, 7]
      ],
      "keypoints": [
        "nose",
        "left_eye",
        "right_eye",
        "left_ear",
        "right_ear",
        "left_shoulder",
        "right_shoulder",
        "left_elbow",
        "right_elbow",
        "left_wrist",
        "right_wrist",
        "left_hip",
        "right_hip",
        "left_knee",
        "right_knee",
        "left_ankle",
        "right_ankle"
      ],
      "id": 1
    }
  ]
}

```

robertklee:
So for evaluation, the model output will be a file containing the predictions for all the test images. I think we'll need to keep track of the image ID so we can reference the ground truth to calculate the OKS/

robertklee:
@julianrocha @nicolepeverley 
So for evaluation, the model output can be a file containing the predictions for all the test images. I think we'll need to either:
- a) keep track of the image ID in the output JSON so we can find and reference the correct ground truth to calculate the OKS/PCK/etc metrics
- b) compute in real time as the model is generating outputs and keep a running tally.

I think a) is a better approach.

PR #9, opened 2021-03-16
Julian/progress report edits

julianrocha:
Main is a protected branch so I cannot push to it. I have proofread all sections and these are the changes I think are necessary.

The only changes are:
- wording/structure
- changed to times new roman


robertklee:
Oops yeah set main to protected to prevent accidental force pushes. You can actually push normally to it. (Only squash or rebase)

PR #10, opened 2021-03-16
Normalize heatmap, add subset argument to train command

julianrocha:
This PR is ready for review. It addresses backlog items 6 and 19. 
- Adds argument to train command to select a subset of the train and val data. Default is 1.0 because we would want to use all data. If you are running experiments and want to go through a bunch of epochs quickly on a smaller set of data, setting subset to 0.1 would use only 10% of the train and 10% of the val. 
- Small fix to heatmaps, normalize value range to [0,1]
- Change data generator test file to be a `.py` file for nice git diffs, but added `# %%` comments which create Jupyter cells in VS Code so you can keep things in memory like a notebook. Best of both worlds really. Shoutout to @robertklee for finding this
- Added the ability to scale heatmaps to a range other than [0,1] for if we want to experiment with placing more emphasis on kp regions. I kept the constant at 1 for now so it has no effect, but in the future, we can easily experiment by varying `HEATMAP_SCALE`


train.py
```diff
@@ -54,6 +54,11 @@ def process_args():
                         type=int,
                         default=DEFAULT_NUM_HG,
                         help='number of hourglass blocks')
+    argparser.add_argument('-sub',
```
robertklee, line 57: can you add some assertions that the value is in range (0,1)? I can see people thinking it is in a % form rather than decimal form.

data_generator_tests.py
```diff
@@ -1,28 +1,94 @@
-
+# %% Load Annotations into dataframes
+import hourglass
+import imp
+imp.reload(hourglass)
```
robertklee, line 4: curious -- what does this do?

hourglass.py
```diff
@@ -36,22 +36,25 @@ def build_model(self, mobile=False, show=False):
...
         print(f"Unfiltered df contains {len(df)} anns")
         df = df.loc[df['is_crowd'] == 0] # drop crowd anns
         df = df.loc[df['num_keypoints'] != 0] # drop anns containing no kps
         df = df.loc[df['bbox_area'] > BBOX_MIN_SIZE] # drop small bboxes
-        print(f"Filtered df contains {len(df)} anns")
         train_df = df.loc[df['source'] == 0]
         val_df = df.loc[df['source'] == 1]
+        if subset != 1.0:
+            train_df = train_df.sample(frac=subset)
+            val_df = val_df.sample(frac=subset)
```
robertklee: personally I think we should keep validation unchanged, because the validation run doesn't actually take that long (less than 2 minutes) since it's not updating any weights. That way, we can compare validation scores across different parameters.

data_generator_tests.py
```diff
@@ -1,28 +1,94 @@
-
+# %% Load Annotations into dataframes
```
robertklee, line 1: Thank you for this script! This will help setup my data aug script and make it a lot easier to debug. :)

hourglass.py
```diff
@@ -87,14 +90,14 @@ def _start_train(self, batch_size, model_base_dir, epochs, initial_epoch, model_
...
+    def train(self, batch_size, model_save_base_dir, epochs, subset):
         current_time = datetime.today().strftime('%Y-%m-%d-%Hh-%Mm')
 
         model_subdir = current_time + '_batchsize_' + str(batch_size) + '_hg_' + str(self.num_stacks)
 
-        self._start_train(batch_size=batch_size, model_base_dir=model_save_base_dir, epochs=epochs, initial_epoch=0, model_subdir=model_subdir, current_time=current_time)
+        self._start_train(batch_size=batch_size, model_base_dir=model_save_base_dir, epochs=epochs, initial_epoch=0, model_subdir=model_subdir, current_time=current_time, subset=subset)
     
-    def resume_train(self, batch_size, model_save_base_dir, model_json, model_weights, init_epoch, epochs, resume_subdir):
+    def resume_train(self, batch_size, model_save_base_dir, model_json, model_weights, init_epoch, epochs, resume_subdir, subset):
```
robertklee, line 99: this may be a problem later, but resuming training and having the model select a different subset might be an issue. But since it's for quick iterations, maybe it's not something to worry about

robertklee (approved):
LGTM. Left a few comments

data_generator_tests.py
julianrocha: force reloading modules allows updates to source file to be reflected in the notebook without needing to restart the notebook kernel. Otherwise module gets loaded once, and trying to import again will just use the module already in memory. This is useful since you wanna get anns in memory (takes 10-30 seconds) by running the first two cells once. But you're gonna be making lots of changes to data generator, so you'll be running the last cell many times and you don't wanna have to load data into memory every time.

hourglass.py
julianrocha: Good point, we definitely cannot use subset for robust/repeatable results. I think saving the subset ids to disk so can resume can use the same set is *probably* overkill for now since this feature is just to make quick tests. I will add comments somewhere so we remember this flaw.

hourglass.py
julianrocha: Hmm, @robertklee do you think it's worth it to have val subset AND train subset arguments? This way if someone so desires, they can perform really fast training with small val as well. If not then I'll just remove the val subset line.

data_generator_tests.py
julianrocha: The example I just gave is more applicable to force reloading data generator on line 60. But I ended up making some changes to `hourglass.py` so that's why I was force reloading it.

hourglass.py
robertklee: I think validation takes so little time that it's probably fine with it removed. Inference is a lot faster typically. But your call though

hourglass.py
robertklee: Yeah, I think it's overkill too since it's just seeing quick improvements to any changes we make. Comments sound good :)

data_generator_tests.py
robertklee: gotcha, thanks!

PR #11, opened 2021-03-19
Add 3 new loss functions, update training metrics to be F1, precision, and recall

robertklee:
We've noticed that training has resulted in the model (likely) finding a local minima of guessing all 0's or a value equal to the mean value of a heat map. This is because the `mean squared error` metric doesn't work well for a large imbalance of 0's and non-0's, which is what our heat map consists of.

## New
- 4 different loss functions: 
    - `keras_mse`: the unmodified mean squared error from keras.losses
    - `weighted_mse`: weighs the 1's as 82 times more intense than the 0's. This parameter may require adjustment.
    - `euclidean_loss`: similar to a mean squared error, but does not involve averaging it. 
    - `focal_loss`: significantly increases the weight of incorrect predictions for the less frequent class (the keypoint location)
- updated metrics to something that is a bit more interpretable, but still not great
    - accuracy just wasn't cutting it for us. This is because it's a classification metric, and is looking at exact matches. I've changed it to `f1`, which is similar to intersection-over-union, which is a _bit_ more useful

The losses are adjustable with command line arguments with the `--loss` flag.

## Quality of Life
- I've added a "notes" flag that allows us to add specific notes to the model session directory name. This is useful if we want to keep track of what we're testing

## TODO
- If someone could test these losses and metrics that would be amazing. See https://stackoverflow.com/questions/50862101/how-to-test-a-custom-loss-function-in-keras

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ opencv-python==4.4.0.42
 matplotlib
 numpy
 requests
+pycocotools @ git+https://github.com/philferriere/cocoapi.git@2929bd2ef6b451054755dfd7ceb09278f935f7ad#subdirectory=PythonAPI
--- a/colab_train.ipynb
+++ b/colab_train.ipynb
@@ -219,10 +219,10 @@
         "outputId": "1f977904-3e1f-4e7d-c0ac-3d8511514c14"
       },
       "source": [
-        "!python3 train.py --epochs 100 --batch 12 --hourglass 4 --subset 1.0\n",
+        "!python3 train.py --epochs 100 --batch 12 --hourglass 4 --subset 1.0 --loss \"euclidean_loss\"\n",
         "\n",
         "# To resume, uncomment the following and point it to the subdirectory. Ensure hourglass matches.\n",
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -8,3 +8,4 @@ opencv-python==4.4.0.42
 matplotlib
 numpy
 requests
+pycocotools @ git+https://github.com/philferriere/cocoapi.git@2929bd2ef6b451054755dfd7ceb09278f935f7ad#subdirectory=PythonAPI
```
julianrocha, line 11: Does this auto-install COCO API for python 3 or is more setup needed? I've setup COCO API on my local but might be useful to know for those who haven't

robertklee: I'm hoping it auto-installs with `pip3 install -r requirements.txt`. This is just what my `pip3 freeze` showed. Not sure if it carries over to new machines

julianrocha (approved):
Great that the loss functions are parameterized especially for this stage of the project. Robert mentioned that these loss functions did not appear to add significant running time in the first tests so looks good to me. I think testing/exploring these loss functions would also be a good idea for the `overfit on one batch` backlog item.

nicolepeverley:
LGTM it'll be nice to test out some different loss functions with the flag 👍

nicolepeverley (approved):

PR #12, opened 2021-03-20
BLOCKED: Convert pipeline to be TPU compatible

julianrocha:
After research and experiments, our model architecture is likely compatible with TPU's on collab, however, our data generator is not. 

The current setup in this PR is able to attach to collab TPUs, compile the model, and start an epoch. However it fails with `Unavailable: {{function_node __inference_train_function_14623}} failed to connect to all addresses`... As per [this](https://github.com/tensorflow/tensorflow/issues/43186),  the error is likely being caused by the data generator as "`tf.keras.utils.Sequence` on TPUs is not supported as it uses py_function underlyingly", this means it must be converted to `tf.data`

As per [this](https://stackoverflow.com/questions/55852831/tf-data-vs-keras-utils-sequence-performance), `tf.data` preprocessing must use TensorFlow operations exclusively. We use `PIL` library for cropping, so converting data generator to `tf.data` would require refactoring of this functionality at least. For now, it appears the swap over, while possible, is likely too much work. `Stacked Hourglass Networks for Human Pose Estimation` paper that we based our model off of was able to train in 3 days on 12GB NVIDIA TitanX GPU.


robertklee:
Adding a few resources here:

https://github.com/tensorflow/tensorflow/issues/39523

PR #13, opened 2021-03-21
Initial Evaluation Visualization

nicolepeverley:
This is an initial PR to predict a set of output heatmaps for one given data generated image.

Example command to run `evaluation.py` (make sure the model json and model weights are in the current directory):

`python evaluation.py --model-json hpe_hourglass_stacks_04_batchsize_012.json --model-weights hpe_epoch39_val_loss_23851522048.0000_train_loss_0.0192.hdf5 --subset 0.1 --hourglass 4`

I've been writing the heatmap images to local disk and overlayed the 17 outputted heatmaps on top of each other into `result.png`, but they tend to cancel each other out, we might want to find a better way of converting this output back into keypoints before we merge them.

### DIFF
--- a/data_generator_tests.py
+++ b/data_generator_tests.py
@@ -76,7 +76,7 @@ import matplotlib.pyplot as plt
 import pylab
 pylab.rcParams['figure.figsize'] = (10.0, 10.0)
 
-from submodules.HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
+from HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
 
 start = time.time()
 X_batch, y_stacked = generator[168]
--- a/evaluation.py
+++ b/evaluation.py
@@ -1,6 +1,8 @@
+# %% Evaluation Imports
 import argparse
 from hourglass import HourglassNet
 from data_generator import DataGenerator
+from submodules.HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
 from constants import *
 from util import *
 
@@ -9,15 +11,48 @@ from datetime import timedelta
 import matplotlib.pyplot as plt
 import numpy as np
 from PIL import Image
+import sys
 import glob
 import cv2
 
-def predict(hourglass, model_json, model_weights, subset):
-    hgnet = HourglassNet(num_classes=NUM_COCO_KEYPOINTS, num_stacks=hourglass, num_channels=NUM_CHANNELS, inres=INPUT_DIM,
+# %% Declare function to horizontally stack images
+# https://stackoverflow.com/questions/30227466/combine-several-images-horizontally-with-python
+def stack_images(images, filename):
+    widths, heights = zip(*(i.size for i in images))
+
+    total_width = sum(widths)
+    max_height = max(heights)
+
+    new_im = Image.new('RGB', (total_width, max_height))
+
+    x_offset = 0
+    for im in images:
+        new_im.paste(im, (x_offset,0))
+        x_offset += im.size[0]
+
+    new_im.save(filename)
+
+# %% Declare function to resize and vertically stack images
+# https://www.geeksforgeeks.org/concatenate-images-using-opencv-in-python/
+def vconcat_resize(img_list, interpolation= cv2.INTER_CUBIC):
+    # take minimum width 
+    w_min = min(img.shape[1] for img in img_list) 
+      
+    # resizing images 
+    im_list_resize = [cv2.resize(img, 
+                      (w_min, int(img.shape[0] * w_min / img.shape[1])), 
+                                 interpolation = interpolation) 
+                      for img in img_list] 
+    # return final image 
+    return cv2.vconcat(im_list_resize) 
+
+# %% Declare function to predict heatmaps
+def predict(hourglass_num, model_json, model_weights, subset):
+    hgnet = HourglassNet(num_classes=NUM_COCO_KEYPOINTS, num_stacks=hourglass_num, num_channels=NUM_CHANNELS, inres=INPUT_DIM,
                             outres=OUTPUT_DIM)
     hgnet._load_model(model_json, model_weights)
 
-    val_df = hgnet.load_and_filter_annotations(DEFAULT_TRAIN_ANNOT_PATH, DEFAULT_VAL_ANNOT_PATH, subset)
+    train_df, val_df = hgnet.load_and_filter_annotations(DEFAULT_TRAIN_ANNOT_PATH, DEFAULT_VAL_ANNOT_PATH, subset)
 
     generator = DataGenerator(
         df=val_df,
@@ -31,39 +66,51 @@ def predict(hourglass, model_json, model_weights, subset):
 
     # Select image to predict heatmaps
     X_batch, y_stacked = generator[168]
-    X = X_batch[0] # take first example of batch
+    y_batch = y_stacked[0] # take first hourglass section
+    X, y = X_batch[0], y_batch[0] # take first example of batch
+
+    image_list = []
+    # Save ground truth heatmaps
+    for i in range(NUM_COCO_KEYPOINTS):
+        heatmap = y[:,:,i]
+        hm = HeatMap(X,heatmap)
+        hm.save(f'ground_truth_heatmap{i}','png', transparency=0.5)
+        image_list.append(f'ground_truth_heatmap{i}.png')
+
+    images = [Image.open(x) for x in image_list]
+    stack_images(images, 'stacked_ground_truth_heatmaps.png')
 
     X = X.reshape(1, 256, 256, 3) # predict needs shape (1, 256, 256, 3)
     output = hgnet.model.predict(X)
     output = np.array(output) # output shape is (hourglass_num, 1, 64, 64, 17)
 
     # Save output heatmaps
-    for i in range(NUM_COCO_KEYPOINTS):
-        plt.figure(i)
-        plt.imsave(f'image{i}.png', output[3, 0, :, :, i])
+    result_list = []
+    for h in range(hourglass_num):
+        image_list = []
+        for i in range(NUM_COCO_KEYPOINTS):
+            plt.figure(i)
+            plt.imsave(f'hourglass{h}_heatmap{i}.png', output[h, 0, :, :, i])
+            image_list.append(f'hourglass{h}_heatmap{i}.png')
 
-    # Overlay heatmaps
-    # https://automaticaddison.com/how-to-blend-multiple-images-using-opencv/
-    # Import all image files with image*.png
-    files = glob.glob ("image*.png")
-    image_data = []
-    for my_file in files:
-        this_image = cv2.imread(my_file, 1)
-        image_data.append(this_image)
+        images = [Image.open(x) for x in image_list]
+        stack_images(images, f'stacked_heatmaps_hourglass{h}.png')
+        result_list.append(f'stacked_heatmaps_hourglass{h}.png')
+
+    result_list.append('stacked_ground_truth_heatmaps.png')
+
+    # Stack each hourglass and ground truth heatmaps into evaluation.png
+    result_imgs = []
+    for x in result_list:
+        result_imgs.append(cv2.imread(x))
+
+    # Resize and vertically stack heatmap images
+    img_v_resize = vconcat_resize(result_imgs) 
     
-    # Calculate blended image
-    dst = image_data[0]
-    for i in range(len(image_data)):
-        if i == 0:
-            pass
-        else:
-            alpha = 1.0/(i + 1)
-            beta = 1.0 - alpha
-            dst = cv2.addWeighted(image_data[i], alpha, dst, beta, 0.0)
- 
-    # Save blended heatmap image
-    cv2.imwrite('result.png', dst)
+    # Save the output heatmaps with ground truth
+    cv2.imwrite('evaluation.png', img_v_resize) 
 
+# %% Declare function to process args
 def process_args():
     argparser = argparse.ArgumentParser(description='Evaluation parameters')
     argparser.add_argument('--model-json',
@@ -98,13 +145,14 @@ def process_args():
 
     return args
 
+# %% Declare main function
 if __name__ == "__main__":
     args = process_args()
 
     print("\n\nModel evaluation start: {}\n".format(time.ctime()))
     evaluation_start = time.time()
 
-    predict(hourglass=args.hourglass, model_json=args.model_json, model_weights=args.model_weights, subset=args.subset)
+    predict(hourglass_num=args.hourglass, model_json=args.model_json, model_weights=args.model_weights, subset=args.subset)
 
     print("\n\nModel evaluation end:   {}\n".format(time.ctime()))
     evaluation_end = time.time()
--- a/evaluation.py
+++ b/evaluation.py
@@ -1,4 +1,4 @@
-from submodules.HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
+from HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
 from constants import *
 
 import matplotlib.pyplot as plt
@@ -32,7 +32,7 @@ class Evaluation():
     def predict_heatmaps(self, h, X):
         h._load_model(self.model_json, self.weights)
 
-        X = X.reshape(1, 256, 256, 3) # predict needs shape (1, 256, 256, 3)
+        X = np.expand_dims(X, axis=0) # add "batch" dimension of 1 because predict needs shape (1, 256, 256, 3)
         predict_heatmaps = h.model.predict(X)
         predict_heatmaps = np.array(predict_heatmaps) # output shape is (num_hg_blocks, 1, 64, 64, 17)
         return predict_heatmaps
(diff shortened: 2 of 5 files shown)


robertklee:
Looks great @nicolepeverley ! yeah overlaying all the heat maps might be an issue. I think we shouldn't stack the _heatmaps_ together yet. At least, not until our model does something. I think the most useful would be a tiling of outputs in one of the following forms:

1. for each output layer (including intermediate layers), generate a set of all 17 outputs in a row (`np.hstack` should do the trick), followed by the corresponding ground truths on the row below (maybe even overlay the ground truth on the original image).
2. similar to 1, but the first row has the ground truths, and then each layer below is the output from layer 0 to layer n (which is the final output). Basically, 1 but vertically stack each row (`np.vstack` I think).

I think (2) would be better since we can see the progression of predictions across each HG stack. @julianrocha thoughts?

The image names ideally should keep track of the epoch information. Easiest way is to probably find some way to reuse the filename since it should be `..._epoch_{:02d}_...`.

evaluation.py
```diff
@@ -0,0 +1,113 @@
...
+from PIL import Image
+import glob
+import cv2
+
+def predict(hourglass, model_json, model_weights, subset):
+    hgnet = HourglassNet(num_classes=NUM_COCO_KEYPOINTS, num_stacks=hourglass, num_channels=NUM_CHANNELS, inres=INPUT_DIM,
+                            outres=OUTPUT_DIM)
+    hgnet._load_model(model_json, model_weights)
+
+    val_df = hgnet.load_and_filter_annotations(DEFAULT_TRAIN_ANNOT_PATH, DEFAULT_VAL_ANNOT_PATH, subset)
```
julianrocha: `load_and_filter_annotations` returns `train_df, val_df` so for now we would want to take the second df. Also, we will probably wanna modify `load_and_filter_annotations` to return 3 dfs, `train_df, val_df, test_df` and we would use `test_df` here. I don't think subset argument is necessary here unless we wanna do evaluations on a subset of the test set.

evaluation.py
```diff
@@ -0,0 +1,113 @@
...
+        num_hg_blocks=DEFAULT_NUM_HG,
+        shuffle=False,  
+        batch_size=1,
+        online_fetch=False)
+
+    # Select image to predict heatmaps
+    X_batch, y_stacked = generator[168]
+    X = X_batch[0] # take first example of batch
+
+    X = X.reshape(1, 256, 256, 3) # predict needs shape (1, 256, 256, 3)
```
julianrocha: We should figure out why this reshaping is necessary. I would have expected the model to be able to take a single image of shape `(256, 256, 3)`. Is the 1 related the batch size that was used when training the model? If so that seems werid.

evaluation.py
```diff
@@ -0,0 +1,113 @@
...
+    output = hgnet.model.predict(X)
+    output = np.array(output) # output shape is (hourglass_num, 1, 64, 64, 17)
```
julianrocha: Another 1 in the shape that confuses me.

evaluation.py
```diff
@@ -0,0 +1,113 @@
...
+    for i in range(len(image_data)):
+        if i == 0:
+            pass
+        else:
+            alpha = 1.0/(i + 1)
+            beta = 1.0 - alpha
+            dst = cv2.addWeighted(image_data[i], alpha, dst, beta, 0.0)
+ 
+    # Save blended heatmap image
+    cv2.imwrite('result.png', dst)
```
julianrocha: More of quality of life comment, but were you able to get any success with matplot working in Collab so that you can view image overlays in a notebook instead of needing to write to disk?

evaluation.py
robertklee: We may just want to `load_and_filter_annotations` to optionally return test, since we don't want to download test2017.zip on the colab VM

evaluation.py
robertklee: yeah I think we need the reshape due to batch size. Although I think `numpy.expand_dims` is better than a reshape here.

julianrocha:
Looks great so far. I agree with @robertklee that there is gonna be quite a bit of work to get a nice plot setup that captures the model output in an understandable way. I think in an ideal world option (2) that Robert suggested would give us the most insight. In all cases, overlaying a heatmap over an image rather than plotting the heatmap by itself would be preferable. Obviously, that's gonna take some time to configure perfectly so we can start with just one joint, or the last hourglass layer and work from there. My only suggestion would be to change the plots to be generated online without needing to write to disk as it would be quicker to try different images but this is contingent on us figuring out the matplot versioning issues that are plaguing some of our local python env setups. That being said, some of these 17kp*(8hg + 1 gt) grids of heatmap plots are probably gonna be massive, so an image file that can be zoomed on may be easier to work with. Your call.

evaluation.py
nicolepeverley: So I tried testing this with batch size not equal to 1 and it still needs the shape to be (1, 256, 256, 3) for some reason.

evaluation.py
nicolepeverley: So I'm currently able to view all plots that are shown in Collab/when saved to Drive but not locally unless saved to disk, I'm not sure if everyone else is having the same issues.

evaluation.py
julianrocha: @nicolepeverley have you tried on different models that have been trained with different batch sizes? In all cases is it always 1?

robertklee: It's because the keras model can process multiple images at once [Source](https://stackoverflow.com/questions/50723287/meaning-of-batch-size-in-model-evaluate):
> The evaluate function of Model has a batch size just in order to speed-up evaluation, as the network can process multiple samples at a time, and with a GPU this makes evaluation much faster. I think the only way to reduce the effect of this would be to set batch_size to one.

evaluation.py
julianrocha: If it ends up being easier, you can always try moving the visualization logic to a notebook. That's what I did for `coco-explorer` And just work in google collab

evaluation.py
julianrocha: Oh perfect that explains it then ty

robertklee: (I didn't know this until now haha)

evaluation.py
julianrocha: Nvm makes sense as per Roberts batch evaluation comment

evaluation.py
robertklee: If you've used the `# %%` short form for a normal `.py` file, you can right click on the file in VS code and convert to a python notebook too

nicolepeverley:
Stacked output of each hourglass set of heatmaps and ground truth heatmaps for a given image:

evaluation

nicolepeverley:
The last commit stacks the output for better visualization, I'm going to check to see if these images need to be saved to disk to accomplish this or not 👍 @julianrocha

wanze-zhang:
Looks great!!!

nicolepeverley:
@Airborne95 and I discussed evaluation moving forward and we're going to update the architecture to look similar to data_generator.py and data_generator_tests.py stay tuned!

Airborne95 (changes requested):
Per my discussion with @nicolepeverley, this file will be refactored into a class similar to `data_generator.py`. This pattern will make it easier to run evaluations from any type of notebook.

evaluation.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __init__(self, model, weights, df, num_hg_blocks, batch_size=1):
+        self.model = model              # json of model to be evaluated
+        self.weights = weights          # weights of model to be evaluated
+        self.df = df                    # df of the the annotations we want
+        self.num_hg_blocks = num_hg_blocks
+        self.batch_size = batch_size
+
+    # https://stackoverflow.com/questions/30227466/combine-several-images-horizontally-with-python
+    def hstack_images(self, images, filename):
```
Airborne95: **Nitpick**
@nicolepeverley I think you mentioned this is a private method so you could change the name to `__hstack_images`.

evaluation.py
```diff
@@ -0,0 +1,101 @@
...
+
+        x_offset = 0
+        for im in images:
+            new_im.paste(im, (x_offset,0))
+            x_offset += im.size[0]
+
+        new_im.save(filename)
+
+    # https://www.geeksforgeeks.org/concatenate-images-using-opencv-in-python/
+    def vstack_images(self, img_list, interpolation=cv2.INTER_CUBIC):
```
Airborne95: **Nitpick**
@nicolepeverley I think you mentioned this is a private method so you could change the name to `__vstack_images`.

Airborne95 (approved):
Its looking really good! I think you did a great job refactoring it into a class 🚀

evaluation.py
robertklee: Good call. I'd avoid `__` prefixes though because double underscores have a special meaning in python. Convention is single underscore `_` to hint private methods/variables

robertklee: > The use of double underscore (__) in front of a name (specifically a method name) is not a convention; it has a specific meaning to the interpreter.

> Python mangles these names and it is used to avoid name clashes with names defined by subclasses.

> This is also called name mangling — the interpreter changes the name of the variable in a way that makes it harder to create collisions when the class is extended later.

evaluation.py
```diff
@@ -0,0 +1,101 @@
...
+
+import matplotlib.pyplot as plt
+import numpy as np
+from PIL import Image
+import cv2
+
+class Evaluation():
+
+    def __init__(self, model, weights, df, num_hg_blocks, batch_size=1):
+        self.model = model              # json of model to be evaluated
```
robertklee: nit: calling this `model` is a bit confusing. Is it the `model_arch` or the model class itself (aka an instance of Hourglass class)?

run_evaluation.py
```diff
@@ -0,0 +1,52 @@
...
+X, y = X_batch[0], y_batch[0] # take first example of batch
+
+predict_heatmaps=eval.predict_heatmaps(h, X)
+print("Received predicted heatmaps")
+stacked_predict_heatmaps=eval.save_stacked_predict_heatmaps(predict_heatmaps)
+print("Saved predicted heatmaps")
+stacked_ground_truth_heatmaps=eval.save_stacked_ground_truth_heatmaps(X, y)
+print("Saved ground truth heatmaps")
+eval.save_stacked_evaluation_heatmaps(stacked_predict_heatmaps, stacked_ground_truth_heatmaps)
+print("Saved stacked evaluation heatmaps")
```
robertklee: Is it possible to generate the fully stacked ones without requiring the other ones to be saved to disk? A few reasons, 1) mainly so it's faster to run as everything is kept in memory, and 2) I read that Google Colab has resource limits on how much you can write to and from Google Drive. They penalize small file read/writes more than big ones.

run_evaluation.py
```diff
@@ -0,0 +1,52 @@
...
+            output_dim=OUTPUT_DIM,
+            num_hg_blocks=DEFAULT_NUM_HG,
+            shuffle=False,  
+            batch_size=1,
+            online_fetch=False)
+
+# Select image to predict heatmaps
+X_batch, y_stacked = generator[168] # choose one image for evaluation
+y_batch = y_stacked[0] # take first hourglass section
+X, y = X_batch[0], y_batch[0] # take first example of batch
```
robertklee, line 42: This is great! It looks like once we have the single image pipeline working, it's a matter of calling 
```
for X_batch, y_stacked in generator:
```

evaluation.py
```diff
@@ -0,0 +1,100 @@
+from submodules.HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
+from constants import *
+
+import matplotlib.pyplot as plt
+import numpy as np
+from PIL import Image
+import cv2
+
+class Evaluation():
```
robertklee, line 8: I'm loving this class refactor!

robertklee (approved):

run_evaluation.py
robertklee: I doubt that this would trigger the resource limits, but if we're running a script that generates evaluation it for a ton of epochs and a ton of different model training sessions, we might want to play it safe. Also, if we end up calling this in the `on_epoch_end` callback, we don't want to be crashing in the middle of training.

nicolepeverley: I'll check on this! I tried using the numpy arrays before and ran into some issues but I'll double check cause it would be much faster 👍

robertklee: if you get all of the hg predictions in a list or np array, you can use `np.hstack([list here])` :)

nicolepeverley: Yeah np.hstack and np.vstack work great for the output heatmaps, but the ground truth heatmaps are using https://github.com/LinShanify/HeatMap so I'm trying to see if I can convert them to np arrays!

robertklee: Sweet! There might be a method in one of the image processing libraries that act as a file buffer to write pictures to. Worst case, since they don't change between layers, we could generate it once and use everywhere too.

evaluation.py
```diff
@@ -0,0 +1,100 @@
...
+        return cv2.vconcat(im_list_resize) 
+
+    # Returns np array of predicted heatmaps for a given image and model
+    def predict_heatmaps(self, h, X):
+        h._load_model(self.model_json, self.weights)
+
+        X = X.reshape(1, 256, 256, 3) # predict needs shape (1, 256, 256, 3)
+        predict_heatmaps = h.model.predict(X)
+        predict_heatmaps = np.array(predict_heatmaps) # output shape is (num_hg_blocks, 1, 64, 64, 17)
+        return predict_heatmaps
```
julianrocha, line 38: Succinct methods are a beauty to read nicely done.

run_evaluation.py
julianrocha: I agree with @robertklee  about only saving the final fully stacked image to disk for simplicity and efficiency. @nicolepeverley if the HeatMap submodule is feeling restrictive, keep in mind we have the option to customize it, so we could add a flag to the plot method that just returns the raw NumPy array instead of trying to plot it (just one idea).

julianrocha: It looks like you were considering this already

evaluation.py
```diff
@@ -0,0 +1,100 @@
...
+        heatmaps.append(stacked_ground_truth_heatmaps)
+        # Stack each hourglass and ground truth heatmaps into heatmap_evaluation.png
+        heatmap_imgs = []
+        for x in heatmaps:
+            heatmap_imgs.append(cv2.imread(x))
+
+        # Resize and vertically stack heatmap images
+        img_v_resize = self._vstack_images(heatmap_imgs) 
+        
+        cv2.imwrite('heatmap_evaluation.png', img_v_resize)
```
julianrocha: Curious what the final use case is for this evaluator. Will we always be trying to plot one and only one image? If we wanna one day plot many, then filenames will need some kind of ID for what image they belong to (annotation ID maybe?). Might be fine if the other files get overwritten since they are intermediate results, but this final result will need an ID. That being said if we only ever intend to use this on one image at a time then this is fine.

julianrocha (approved):
Most of my comments were related to how files are handled. Otherwise, this is great, gonna give us huge insight into how the model handles specific images and how the heatmaps change across hourglasses (<- useful for picking the best hg number). Also a huge shoutout to the refactoring that happened here, the difference in readability was night and day compared to the first time I read this PR so great stuff.

evaluation.py
```diff
@@ -0,0 +1,77 @@
...
+                                    interpolation = interpolation) 
+                        for img in img_list] 
+        # return final image 
+        return cv2.vconcat(im_list_resize) 
+
+    # Returns np array of predicted heatmaps for a given image and model
+    def predict_heatmaps(self, h, X):
+        h._load_model(self.model_json, self.weights)
+
+        X = X.reshape(1, 256, 256, 3) # predict needs shape (1, 256, 256, 3)
```
robertklee: small nit: 
```suggestion
        X = np.expand_dims(X, axis=0) # add "batch" dimension of 1 because predict needs shape (1, 256, 256, 3)
```

evaluation.py
```diff
@@ -0,0 +1,77 @@
...
+                stacked_hourglass_heatmaps = np.array(stacked_predict_heatmaps)
+            else:
+                stacked_hourglass_heatmaps = np.vstack((stacked_hourglass_heatmaps, stacked_predict_heatmaps))
+        return stacked_hourglass_heatmaps
+
+    #  Saves stacked predicted heatmaps and stacked ground truth heatmaps in one evaluation image
+    def save_stacked_evaluation_heatmaps(self, stacked_predict_heatmaps_file, stacked_ground_truth_heatmaps_file, filename):
+        heatmap_imgs = []
+        heatmap_imgs.append(cv2.imread(stacked_predict_heatmaps_file))
+        heatmap_imgs.append(cv2.imread(stacked_ground_truth_heatmaps_file))
```
robertklee, line 81: It looks like we don't need to pass in files anymore, right?

run_evaluation.py
```diff
@@ -0,0 +1,59 @@
...
+stacked_predict_heatmaps=eval.stacked_predict_heatmaps(predict_heatmaps)
+print("Received stacked predicted heatmaps")
+stacked_ground_truth_heatmaps=eval.stacked_ground_truth_heatmaps(X, y)
+print("Received stacked ground truth heatmaps")
+
+# Save stacked images to disk
+plt.imsave('stacked_predict_heatmaps.png', stacked_predict_heatmaps)
+plt.imsave('stacked_ground_truth_heatmaps.png', stacked_ground_truth_heatmaps)
+filename = 'heatmap_evaluation.png'
+eval.save_stacked_evaluation_heatmaps('stacked_predict_heatmaps.png', 'stacked_ground_truth_heatmaps.png', filename)
```
robertklee: Do we need to save still?

run_evaluation.py
```diff
@@ -0,0 +1,59 @@
...
+y_batch = y_stacked[0] # take first hourglass section
+X, y = X_batch[0], y_batch[0] # take first example of batch
+
+# Retrieve prediction and ground truth heatmaps
+predict_heatmaps=eval.predict_heatmaps(h, X)
+print("Received predicted heatmaps")
+stacked_predict_heatmaps=eval.stacked_predict_heatmaps(predict_heatmaps)
+print("Received stacked predicted heatmaps")
+stacked_ground_truth_heatmaps=eval.stacked_ground_truth_heatmaps(X, y)
+print("Received stacked ground truth heatmaps")
```
robertklee: would it be better for `eval` to have one function to call that calls these methods?

robertklee (approved):
Overall looks great!

nicolepeverley:
Okay so the reason I was having trouble vertically stacking/concatenating the predicted heatmaps and ground truth heatmaps  is because the ground truth heatmaps get returned as 3 dimensional np arrays (our predicted heatmaps are 2 dimensional np arrays). I'm not sure how to get around this so for now I've left it that if you want to see the predicted heatmaps and ground truth heatmaps stacked together you'll have to save them!

notebooks/input_pipeline_demo.ipynb
```diff
@@ -990,7 +1022,7 @@
         "import pylab\n",
         "pylab.rcParams['figure.figsize'] = (10.0, 10.0)\n",
         "\n",
-        "from submodules.HeatMap import HeatMap # https://github.com/LinShanify/HeatMap\n",
+        "from HeatMap import HeatMap # https://github.com/LinShanify/HeatMap\n",
```
nicolepeverley, line 1025: This is the only change here for this file!

robertklee (approved):
LGTM! This isn't a blocker but one possible option to temporarily save an image in-memory is [this StackOverflow answer](https://stackoverflow.com/questions/33916409/how-to-save-image-in-memory-and-upload-using-pil)

```
temp = BytesIO()
im.save(temp, format="png")
```

robertklee:
We can explore this tiny QOL improvement after merging though

data_generator_tests.py
```diff
@@ -76,7 +76,7 @@ def display_img(annId):
 import pylab
 pylab.rcParams['figure.figsize'] = (10.0, 10.0)
 
-from submodules.HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
+from HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
```
julianrocha, line 79: ty for fixing the dependencies

julianrocha (approved):
Juicy

PR #15, opened 2021-03-24
Corey/data split

Ckoelewyn:
This PR covers task **Create a new train/val from train (possibly need stratification)**. 

Use the `pickle_maker.ipynb` notebook for creating pickles in your GDrive
Use the `--pickle` argument with `train.py` in `colab_train.ipynb` to specify which pickle you want to train on.

Features
 * df_pickler module with functions for creating pickle files from dataframes and reading data frames from pickle files
   * Creates training, validation, and testing datasets (training and validation come from COCO's training, testing comes from COCO's testing)
 * Notebook file `pickle_maker.ipynb` for creating pickle files in your google drive
 * new arg `--pickle` specifies which pickle you want to consume
    * If pickle flag not used train.py uses `hourglass.load_and_filter_annotations()` to generate training and validation dataframes
    * If the pickle doesn't exist, an error is output and train.py uses `hourglass.load_and_filter_annotations()` to generate training and validation dataframes
  * new script `anno_dl.sh` for downloading only annotation zips.

Potential issues
  * naming scheme for model folders won't reflect which pickle this came from
  * stratification is not currently implemented, could result in training and validation sets having unequal representation of different parameters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,6 @@ dmypy.json
 
 # VS Code Settings
 .vscode/*
+
+#Pickles and pickle directories
+Pickles/*
--- a/df_pickler.py
+++ b/df_pickler.py
@@ -1,0 +1,78 @@
+# Pickles and unpickles data frames for maintaining consistant dataframes.
+
+import numpy as np
+import pandas as pd
+import os
+import skimage.io as io
+import coco_df
+from constants import *
+from pycocotools.coco import COCO
+from sklearn.model_selection import train_test_split
+
+
+
+
+def check_pickle_folder(pickle_folder, pickle_path = DEFAULT_PICKLE_PATH) -> bool:
+    # Checks if the folder exists
+    return os.path.isdir(os.path.join(pickle_path, pickle_folder))
+    
+
+def create_folder(pickle_folder, pickle_path = DEFAULT_PICKLE_PATH):
+    # Creates a folder
+    pf = os.path.join(pickle_path, pickle_folder)
+    if check_pickle_folder(pickle_folder, pickle_path):
+        print(f'Folder {pickle_folder} exists')
+        return pf
+    print(pf)
+    os.mkdir(pf)
+    return pf
+
+def make_pickle(is_crowd: bool, drop_empty: bool, scale: float,TV_split: float, drop_small_bbox=BBOX_MIN_SIZE):
+    if scale <= 0.0 or scale > 1.0:
+        raise ValueError
+    if TV_split <= 0.0 or scale > 1.0:
+        raise ValueError
+    pickle_name = pickle_namer(is_crowd, drop_empty, scale, TV_split, drop_small_bbox)
+    pickle_folder = create_folder(pickle_name)
+    df = coco_df.get_df(DEFAULT_TRAIN_ANNOT_PATH, DEFAULT_VAL_ANNOT_PATH)
+    if is_crowd:
+        df = df.loc[df['is_crowd'] == 0]
+    if drop_empty:
+        df = df.loc[df['num_keypoints'] != 0]
+    df = df.loc[df['bbox_area'] > drop_small_bbox]
+    df_TrV = df.loc[df['source'] == 0] #Data from the training set for splitting into train/val
+    df_Tst = df.loc[df['source'] == 1] #Data from the val set for testing
+    t_size = scale * TV_split
+    v_size = scale * (1-TV_split)
+    df_Tr, df_V = train_test_split(df_TrV, train_size=t_size, test_size=v_size) # We can stratify here if wanted as well
+    df_Tr.to_pickle(os.path.join(pickle_folder, f'Train_{pickle_name}.pkl'))
+    df_V.to_pickle(os.path.join(pickle_folder, f'Val_{pickle_name}.pkl'))
+    df_Tst.to_pickle(os.path.join(pickle_folder, f'Test_{pickle_name}.pkl'))
+    return
+    
+
+def get_pickle(pickle_name):
+    pickle_folder = os.path.join(DEFAULT_PICKLE_PATH, pickle_name)
+    df_Tr = pd.read_pickle(os.path.join(pickle_folder, f'Train_{pickle_name}.pkl'))
+    df_V = pd.read_pickle(os.path.join(pickle_folder, f'Val_{pickle_name}.pkl'))
+    df_Tst = pd.read_pickle(os.path.join(pickle_folder, f'Test_{pickle_name}.pkl'))
+    return df_Tr.reset_index(), df_V.reset_index(), df_Tst.reset_index()
+
+def pickle_namer(is_crowd, drop_empty, scale, TV_split, drop_small_bbox):
+    filename = f''
+    filename += f'C-{is_crowd}_'
+    filename += f'DE-{drop_empty}_'
+    filename += f'S-{int(scale*100)}_'
+    filename += f'TVs-{int(TV_split*100)}_'
+    filename += f'DSBB-{drop_small_bbox}'
+    return filename
+
+def parse_pickle_name(filename):
+    data =  [val for val in filename.split('.')[0].split('_')]
+    print(f'Settings for pickle ({filename}):')
+    print(f'  Crowd enabled              : {data[0].split("-")[1]}')
+    print(f'  Drop Empty                 : {data[1].split("-")[1]}')
+    print(f'  Dataframe scaling          : {data[2].split("-")[1]/100}')
+    print(f'  Test/validation splot      : {data[3].split("-")[1]/100}')
+    print(f'  Bounding box lower limit   : {data[4].split("-")[1]}')
+    return
--- a/hourglass.py
+++ b/hourglass.py
@@ -13,6 +13,7 @@ import coco_df
 from constants import *
 from data_augmentation import *
 from data_generator import DataGenerator
+from df_pickler import get_pickle
 from hourglass_blocks import (bottleneck_block, bottleneck_mobile,
                               create_hourglass_network)
 from loss import *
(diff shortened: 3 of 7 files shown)


hourglass.py
```diff
@@ -53,10 +54,13 @@ def load_and_filter_annotations(self, path_to_train_anns,path_to_val_anns, subse
         print(f"Train/Val dfs contains {len(train_df)}/{len(val_df)} anns")
         return train_df.reset_index(), val_df.reset_index()
 
-    def _start_train(self, batch_size, model_base_dir, epochs, initial_epoch, model_subdir, current_time, subset):
+    def _start_train(self, batch_size, model_base_dir, epochs, initial_epoch, model_subdir, current_time, subset, pickle_name):
         self._compile_model()
-
-        train_df, val_df = self.load_and_filter_annotations(DEFAULT_TRAIN_ANNOT_PATH, DEFAULT_VAL_ANNOT_PATH, subset)
+        # insert logic here for pickled dataframes
+        if pickle_name != None:
```
julianrocha: Use `is not` instead of `!=`
"Comparisons to singletons like None should always be done with is or is not, never the equality operators."
[PEP 8 Style Guide](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

df_pickler.py
```diff
@@ -0,0 +1,78 @@
...
+    df = df.loc[df['bbox_area'] > drop_small_bbox]
+    df_TrV = df.loc[df['source'] == 0] #Data from the training set for splitting into train/val
+    df_Tst = df.loc[df['source'] == 1] #Data from the val set for testing
+    t_size = scale * TV_split
+    v_size = scale * (1-TV_split)
+    df_Tr, df_V = train_test_split(df_TrV, train_size=t_size, test_size=v_size) # We can stratify here if wanted as well
+    df_Tr.to_pickle(os.path.join(pickle_folder, f'Train_{pickle_name}.pkl'))
+    df_V.to_pickle(os.path.join(pickle_folder, f'Val_{pickle_name}.pkl'))
+    df_Tst.to_pickle(os.path.join(pickle_folder, f'Test_{pickle_name}.pkl'))
+    return
```
julianrocha, line 51: These are hard-coded filtering rules like we have in `load_and_filter_annotations`. I think one of the benefits we get with pickling is the ability to filter/partition the dataset as we want, and can easily make new pickles for different types of splits/filters. For this reason, I would argue that the filtering rules applied here should be moved to the responsibility of the user of `pickle_maker.ipynb`. So the user can create a `coco_df`, and filter what they want into 3 dfs, and then could pass all 3 to `make_pickle`. This is very much a question of requirements, I'm suggesting this logic is moved to the notebook, but I'm curious what others think as this currently offers the benefit of documenting our filters/partitions in code rather than hoping that whoever created the pickle filtered correctly.

df_pickler.py
```diff
@@ -0,0 +1,78 @@
...
+    return df_Tr.reset_index(), df_V.reset_index(), df_Tst.reset_index()
+
+def pickle_namer(is_crowd, drop_empty, scale, TV_split, drop_small_bbox):
+    filename = f''
+    filename += f'C-{is_crowd}_'
+    filename += f'DE-{drop_empty}_'
+    filename += f'S-{int(scale*100)}_'
+    filename += f'TVs-{int(TV_split*100)}_'
+    filename += f'DSBB-{drop_small_bbox}'
+    return filename
```
julianrocha, line 68: Should we decide to move filtering logic to the notebook level, then we would likely need to pass custom filenames to `make_pickle` instead of having a `pickle_namer`

train.py
```diff
@@ -83,6 +85,9 @@ def process_args():
     argparser.add_argument('--resume-subdir',
                         default=None,
                         help='Subdirectory containing architecture json and weights')
+    argparser.add_argument('--pickle',
+                        default=None,
+                        help='Name of folder with pickled dataframes')
```
julianrocha, line 94: It might be nice to merge some basic pickles into the repository so that everyone has the same set of `train/val/test` to work with. Obviously people could create new ones but a basic set to get started is nice to quickly clone a new repo and start training without needing to download anything else. How big are these files btw? (not sure if we wanna avoid checking big files into the repo)

hourglass.py
```diff
@@ -53,10 +54,13 @@ def load_and_filter_annotations(self, path_to_train_anns,path_to_val_anns, subse
...
+            train_df, val_df, test_df = get_pickle(pickle_name)
```
julianrocha, line 64: Since `test_df` likely won't be used in this method, could change the unpacking to: `train_df, val_df, _`

scripts/anno_dl.sh
```diff
@@ -0,0 +1,58 @@
+# Adapted from coco_dl.sh
+# Removed the image datasets, they are not required for creating dataframe objects
```
julianrocha: noice addition

robertklee: I'm wondering if this script can be merged with the existing `coco_dl.sh` script. It seems all we need to add is another command line argument after dataset path. (From first glance I don't see anything new in this script except for the removal of the image zip download)

robertklee:
I think you accidentally deleted `submodules/HeatMap`

pickle_maker.ipynb
```diff
@@ -0,0 +1 @@
...loss function matches across training sessions"],"execution_count":null,"outputs":[{"output_type":"stream","text":["WARNING:tensorflow:From /tensorflow-1.15.2/python3.7/keras/backend/tensorflow_backend.py:422: The name tf.global_variables is deprecated. Please use tf.compat.v1.global_variables instead.\n","\n","WARNING:tensorflow:From /tensorflow-1.15.2/python3.7/keras/callbacks/tensorboard_v1.py:200: The name tf.summary.merge_all is deprecated. Please use tf.compat.v1.summary.merge_all instead.\n","\n","WARNING:tensorflow:From /tensorflow-1.15.2/python3.7/keras/callbacks/tensorboard_v1.py:203: The name tf.summary.FileWriter is deprecated. Please use tf.compat.v1.summary.FileWriter instead.\n","\n","2021-03-23 08:17:48.371388: I tensorflow/stream_executor/platform/default/dso_loader.cc:44] Successfully opened dynamic library libcudnn.so.7\n","2021-03-23 08:17:49.616987: I tensorflow/stream_executor/platform/default/dso_loader.cc:44] Successfully opened dynamic library libcublas.so.10\n","WARNING:tensorflow:From /tensorflow-1.15.2/python3.7/keras/callbacks/tensorboard_v1.py:343: The name tf.Summary is deprecated. Please use tf.compat.v1.Summary instead.\n","\n"],"name":"stdout"}]}]}
```
julianrocha: <p>Duplicated</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/robertklee/COCO-Human-Pose/pull/15/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

julianrocha: <p>The <code>make_pickle()</code>  looks nice to use. If we decide to move logic to the notebook, then we can create <code>coco_df</code>  here , filter it and split it into <code>train,val,test</code>  dfs and pass them along with the filename to <code>make_pickle()</code>  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/robertklee/COCO-Human-Pose/pull/15/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

julianrocha: <p>This train cell should probs be removed. The <code>--pickle</code>  flag for training should be moved to <code>colab-train.ipynb</code>  so that people know how to use it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/robertklee/COCO-Human-Pose/pull/15/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

.gitignore
```diff
@@ -132,4 +132,7 @@ dmypy.json
 .pyre/
 
 # VS Code Settings
-.vscode/*
\ No newline at end of file
+.vscode/*
+
+#Pickles and pickle directories
+Pickles/*
```
robertklee: nit: end with newline character for better diffs

julianrocha (commented):
Looks great. Left some comments, we should clarify if the suggestion I made is what we want. Also, this PR removes the HeatMap Submodule which we should not do (since we still need this submodule)

hourglass.py
robertklee: we may need some way to load test (which for us might be the val2017 set) separately than the others, since we'll be running evaluation on the test set. We'll have to decide how many entries to set aside for validation and testing.

pickle_maker.ipynb
robertklee: I think for clearer diffs, could you remove the outputs and throw the JSON in a JSON prettifier?

robertklee:
Overall it looks like a great addition for some much needed data splitting options for us. I haven't fully reviewed it (I'm working on the assignment) but I've left a few comments on things I noticed. Thanks for knocking this off our backlog.

train.py
Ckoelewyn: Full set of Train/Val/Test with no scaling is 151MB

robertklee: We could deploy them via GitHub Package Releases. Up to 1 GB per file, and I believe it also generates a link we can use in our scripts. What was the downside to doing the splitting in-memory? Was it so we have consistent train/val splits on the training set (i.e. no chance of a different split across training sessions)? Could we just save a csv of annotation IDs, and populate the data frames using those IDs?

Ckoelewyn: Consistency between the sets. If we get ejected from colab deep into a training set, having the splits saved would allow training to resume with identical training/validation/testing as before. If we wanted to train for multiple days across the same dataset we no longer need to worry about the 24hr timeout. 

Saving it as a Csv is doable, but It’s [much slower than pickle](https://towardsdatascience.com/the-best-format-to-save-pandas-data-414dca023e0d?gi=c431087feed5)

I’ll look into GitHub package releases tomorrow.

robertklee: Sounds good, thanks for the clarification. Fully agree with your analysis. Then I think GitHub package releases might be the best option for us. We would be limited to 2 GB per 'release' which sounds like it's plenty of breathing room for us. We'd need to confirm that the package URL stays constant.

scripts/anno_dl.sh
Ckoelewyn: Per our discussion this afternoon I've wrapped the image downloading portion in `coco_dl.sh`. Passing `no_image` as the second argument prevents the script from downloading the image dataset.

robertklee (approved):
LGTM!

PR #16, opened 2021-03-26
Prepare for batch Evaluation

julianrocha:
This PR is trying to lay the groundwork for batch evaluation of the models we have. This includes visualizing a batch of representative images for the set of models we have, as well as eventually performing OKS on the test set on a batch of models. A lot of the OKS work depends on the output pipeline being finished, so for now that is on hold. 

Here is what this PR has done:
- Load model at eval init stage for re-use
- Pass model paths to load any model from anywhere
- Stop writing intermediate files
- Write visualization files at `output/model_sub_dir/epoch_num/ann_id.png` to ensure they are unique and can be associated with the model that generated them.
- Perform evaluation on a batch of images rather than one at a time
- Generated representative set 10 images from the validation set, stored in `representative_set.pkl`. This df has been checked into the repo since it's small so that anyone can visualize it. It consists of `ann_ids = [442619, 559508, 560228, 220340, 227893, 1735793, 477234, 1731118, 476220, 464797]` We can add to/modify this set if we want, but keeping it small reduces run time. 
- Added colour injection to heatmaps for the contrast of activations.
- Made new `run_visualizaton.ipynb` to run visualization on a model sub_dir across epochs.


evaluation.py
```diff
@@ -4,15 +4,17 @@
...
-        self.weights = weights          # weights of model to be evaluated
-        self.df = df                    # df of the the annotations we want
-        self.num_hg_blocks = num_hg_blocks
-        self.batch_size = batch_size
+    def __init__(self, sub_dir, model_json, weights, h_net): 
+        self.model_json = os.path.join(sub_dir,model_json)         # json of model to be evaluated
+        self.weights = os.path.join(sub_dir,weights)          # weights of model to be evaluated
+        self.num_hg_blocks = int(re.match(r'.*stacks_(\d\d)_.*',model_json).group(1))
+        h_net._load_model(self.model_json, self.weights)
+        self.model = h_net.model
```
julianrocha: Add sub dir for where to load model from. Grab num hg from model file name, store model as instance variable so it gets loaded once at `Evaluation` instance initialization.

evaluation.py
```diff
@@ -29,11 +31,9 @@ def _vstack_images(self, img_list, interpolation=cv2.INTER_CUBIC):
...
     # Returns np array of predicted heatmaps for a given image and model
-    def predict_heatmaps(self, h, X):
-        h._load_model(self.model_json, self.weights)
-
+    def predict_heatmaps(self, X):
         X = np.expand_dims(X, axis=0) # add "batch" dimension of 1 because predict needs shape (1, 256, 256, 3)
-        predict_heatmaps = h.model.predict(X)
+        predict_heatmaps = self.model.predict(X)
         predict_heatmaps = np.array(predict_heatmaps) # output shape is (num_hg_blocks, 1, 64, 64, 17)
         return predict_heatmaps
```
julianrocha: Moved the model loading to instance evaluation. On my laptop loading a model from disk takes ~10 seconds.

evaluation.py
```diff
@@ -66,19 +66,21 @@ def stacked_predict_heatmaps(self, predict_heatmaps):
...
-
-        # Save stacked images to disk
-        plt.imsave(stacked_predict_heatmaps_file, stacked_predict_heatmaps)
-        plt.imsave(stacked_ground_truth_heatmaps_file, stacked_ground_truth_heatmaps)
+        
+        # Reshape heatmaps to 3 channels, normalize channels to [0,255]
+        stacked_predict_heatmaps = cv2.cvtColor(stacked_predict_heatmaps, cv2.COLOR_GRAY2RGB)
+        stacked_predict_heatmaps = cv2.normalize(stacked_predict_heatmaps, None, alpha=0, beta=255.0, norm_type=cv2.NORM_MINMAX, dtype=cv2.CV_32F)
+        stacked_ground_truth_heatmaps = cv2.cvtColor(stacked_ground_truth_heatmaps, cv2.COLOR_BGRA2RGB)
+        stacked_ground_truth_heatmaps = cv2.normalize(stacked_ground_truth_heatmaps, None, alpha=0, beta=255.0, norm_type=cv2.NORM_MINMAX, dtype=cv2.CV_32F)
```
julianrocha: Heatmaps come in one channel, cv2 needs 3. GT comes in 4 channels (including alpha), we need 3. Heatmaps come in a wide range of values [negative, positive], we need [0,255]. GT also needs to be in [0,255]

run_evaluation.py
```diff
@@ -3,32 +3,32 @@
 import imp
 imp.reload(hourglass)
 from hourglass import HourglassNet
-import data_generator
-imp.reload(data_generator)
-from data_generator import DataGenerator
-import evaluation
-imp.reload(evaluation)
-from evaluation import Evaluation
```
julianrocha, line 11: Moved reloads to the cells that run quicker

evaluation.py
```diff
@@ -4,15 +4,17 @@
...
-    def __init__(self, model_json, weights, df, num_hg_blocks, batch_size=1):
-        self.model_json = model_json              # json of model to be evaluated
-        self.weights = weights          # weights of model to be evaluated
-        self.df = df                    # df of the the annotations we want
-        self.num_hg_blocks = num_hg_blocks
-        self.batch_size = batch_size
+    def __init__(self, sub_dir, model_json, weights, h_net): 
+        self.model_json = os.path.join(sub_dir,model_json)         # json of model to be evaluated
+        self.weights = os.path.join(sub_dir,weights)          # weights of model to be evaluated
+        self.num_hg_blocks = int(re.match(r'.*stacks_(\d\d)_.*',model_json).group(1))
```
robertklee: We may want to pass in the base directory too if we want to run on all trained models. We can pass in an empty string `''` for `sub_dir` if the model files are saved directly in the model folder. `os.path.join` will omit any strings which are empty. (It doesn't handle `None` though)

```suggestion
    def __init__(self, base_dir, sub_dir, model_json, weights, h_net): 
        self.model_json = os.path.join(base_dir, sub_dir, model_json)         # json of model to be evaluated
        self.weights = os.path.join(base_dir, sub_dir, weights)          # weights of model to be evaluated
        self.num_hg_blocks = int(re.match(r'.*stacks_(\d\d)_.*',model_json).group(1))
```

robertklee (approved):
Overall this is a solid groundwork for future eval work. Nice!

evaluation.py
```diff
@@ -66,19 +66,21 @@ def stacked_predict_heatmaps(self, predict_heatmaps):
...
+        stacked_predict_heatmaps = cv2.normalize(stacked_predict_heatmaps, None, alpha=0, beta=255.0, norm_type=cv2.NORM_MINMAX, dtype=cv2.CV_32F)
+        stacked_ground_truth_heatmaps = cv2.cvtColor(stacked_ground_truth_heatmaps, cv2.COLOR_BGRA2RGB)
+        stacked_ground_truth_heatmaps = cv2.normalize(stacked_ground_truth_heatmaps, None, alpha=0, beta=255.0, norm_type=cv2.NORM_MINMAX, dtype=cv2.CV_32F)
         filename = filename
         
         heatmap_imgs = []
-        heatmap_imgs.append(cv2.imread(stacked_predict_heatmaps_file))
-        heatmap_imgs.append(cv2.imread(stacked_ground_truth_heatmaps_file))
+        heatmap_imgs.append(stacked_predict_heatmaps)
+        heatmap_imgs.append(stacked_ground_truth_heatmaps)
```
nicolepeverley, line 108: This is great that we can store the stacked heatmaps in memory now 👍  Do we still want to write the resulting vertically stacked image to disk or should we return it to possibly just display/use in `run_evaluation.py`?

nicolepeverley (approved):
Awesome improvement!! I think this will be super helpful for using the evaluation between epochs. I just left a note about saving the final image but otherwise looks good to me 💯

evaluation.py
julianrocha: I think to disk. Since we will wanna run this on possibly many images and possibly many models. Tomorrow I will look at organizing/setting up unique fnames for these eval photos, as well as evaluating a whole batch of images at once.

evaluation.py
julianrocha: done yo

evaluation.py
```diff
@@ -48,11 +48,10 @@ def predict_heatmaps(self, X):
         predict_min = np.min(predict_heatmaps)
         predict_var = np.var(predict_heatmaps.flatten())
         print('Mean: {:0.6e}\t Max: {:e}\t Min: {:e}\t Variance: {:e}'.format(predict_mean, predict_max, predict_min, predict_var))
+        # TODO: is this how we want to normalize output? Should we instead enforce model output to be normalized?
```
robertklee: This is temporary for now. I'll add a sigmoid layer at the end of each model output and see how it does.

evaluation.py
```diff
@@ -89,11 +88,13 @@ def save_stacked_evaluation_heatmaps(self, X, y, filename):
         stacked_predict_heatmaps=self.stacked_predict_heatmaps(predict_heatmaps)
         stacked_ground_truth_heatmaps=self.stacked_ground_truth_heatmaps(X, y)
         
-        # Reshape heatmaps to 3 channels, normalize channels to [0,255]
-        stacked_predict_heatmaps = cv2.cvtColor(stacked_predict_heatmaps, cv2.COLOR_GRAY2RGB)
-        stacked_predict_heatmaps = cv2.normalize(stacked_predict_heatmaps, None, alpha=0, beta=255.0, norm_type=cv2.NORM_MINMAX, dtype=cv2.CV_32F)
+        # Reshape heatmaps to 3 channels with colour injection, normalize channels to [0,255]
+        # TODO - normalize individual heatmaps rather than a whole stack at once.
```
robertklee: This looks great! Re: normalizing individual heat maps, we might not want to do that because it'll just scale heat maps where it predicts nothing (very small values) to everything is lit up. we should set max across an entire hg layer or across the entire model imo

robertklee (approved):
Looks great, left a few comments

evaluation.py
robertklee: Batching would be amazing. I'm 3 hours in and halfway through. Even though I'm running on pure CPU, it's definitely not a slouch normally.

evaluation.py
julianrocha: Good thinking. No change necessary then

run_evaluation.py
```diff
@@ -3,49 +3,53 @@
...
 from constants import *
 import matplotlib.pyplot as plt
+import os
```
robertklee, line 8: Just making a few suggestions to add a print for start and end time 
```suggestion
import os
import time
```

run_evaluation.py
```diff
@@ -3,49 +3,53 @@
...
+imp.reload(HeatMap)
+
+subdir = '2021-03-22-20h-23m_batchsize_12_hg_8_loss_weighted_mse_aug_medium_resume_2021-03-25-20h-02m'
+
+eval = evaluation.Evaluation(
+    base_dir=DEFAULT_MODEL_BASE_DIR,
+    sub_dir=subdir,
+    epoch=43,
+    h_net=h)
 print("Created Evaluation instance")
```
robertklee, line 25:
```suggestion

print("\n\nEval start:   {}\n".format(time.ctime()))

```

run_evaluation.py
```diff
@@ -3,49 +3,53 @@
...
 X, y = X_batch[0], y_batch[0] # take first example of batch
-
+plt.imshow(X)
 # Save stacked heatmap images to disk
-stacked_predict_heatmaps_file = 'stacked_predict_heatmaps.png'
-stacked_ground_truth_heatmaps_file = 'stacked_ground_truth_heatmaps.png'
 filename = 'heatmap_evaluation.png'
-eval.save_stacked_evaluation_heatmaps(h, X, y, stacked_predict_heatmaps_file, stacked_ground_truth_heatmaps_file, filename)
+eval.save_stacked_evaluation_heatmaps(X, y, os.path.join(DEFAULT_OUTPUT_BASE_DIR, filename))
 print(f"Saved stacked evaluation heatmaps as {filename} to disk")
```
robertklee, line 70:
```suggestion

print("\n\nEval end:   {}\n".format(time.ctime()))

```

run_evaluation.py
```diff
@@ -3,49 +3,50 @@
...
-    batch_size=1)
-print("Created Evaluation instance")
+import pandas as pd
+import evaluation
+import HeatMap
+imp.reload(evaluation)
+imp.reload(HeatMap)
+
+representative_set_df = pd.read_pickle(os.path.join(DEFAULT_PICKLE_PATH, 'representative_set.pkl'))
+subdir = '2021-03-22-20h-23m_batchsize_12_hg_8_loss_weighted_mse_aug_medium_resume_2021-03-25-20h-02m'
```
robertklee, line 21: get ready for these to double in length lol

HeatMap.py
```diff
@@ -97,4 +97,5 @@ def get_heatmap_array(self,transparency=0.7,color_map='bwr'):
         fig.canvas.draw()
         #Turn heatmap into numpy array
         heatmap_array = np.array(fig.canvas.renderer._renderer)
+        plt.close('all')
```
robertklee, line 100: good catch

evaluation.py
```diff
@@ -1,20 +1,38 @@
...
-        self.model_json = model_json              # json of model to be evaluated
-        self.weights = weights          # weights of model to be evaluated
-        self.df = df                    # df of the the annotations we want
-        self.num_hg_blocks = num_hg_blocks
-        self.batch_size = batch_size
+    def __init__(self, model_sub_dir, epoch, model_base_dir=DEFAULT_MODEL_BASE_DIR, output_base_dir=DEFAULT_OUTPUT_BASE_DIR): 
+        # automatically retrieve json and weights
+        self.model_sub_dir=model_sub_dir
+        self.epoch=epoch
+        match = re.match(r'(.*)(_resume_.*$)', model_sub_dir)
```
robertklee, line 22: nice. this will clean up our output folder considerably

evaluation.py
```diff
@@ -31,31 +49,30 @@ def _vstack_images(self, img_list, interpolation=cv2.INTER_CUBIC):
...
-        normalized_heatmaps = predict_heatmaps / predict_max
-        normalized_heatmaps = normalized_heatmaps - predict_min
-        normalized_heatmaps = predict_heatmaps / np.max(predict_heatmaps)
-
-        return normalized_heatmaps
+    def predict_heatmaps(self, X_batch):
+        predicted_heatmaps = np.array(self.model.predict(X_batch)) # output shape is (num_hg_blocks, X_batch_size, 64, 64, 17)
+        for i in range(len(X_batch)):
+            hm = predicted_heatmaps[:,i,]
+            print('model prediction metrics: ')
```
robertklee: I know I added this, but I'd remove for now. It ended up cluttering up my console and slowing down the code

evaluation.py
```diff
@@ -31,31 +49,30 @@ def _vstack_images(self, img_list, interpolation=cv2.INTER_CUBIC):
...
+            predict_mean = np.mean(hm)
+            predict_max = np.max(hm)
+            predict_min = np.min(hm)
+            predict_var = np.var(hm.flatten())
+            print('Mean: {:0.6e}\t Max: {:e}\t Min: {:e}\t Variance: {:e}'.format(predict_mean, predict_max, predict_min, predict_var))
```
robertklee: I know I added this, but I'd remove for now. It ended up cluttering up my console and slowing down the code

evaluation.py
```diff
@@ -31,31 +49,30 @@ def _vstack_images(self, img_list, interpolation=cv2.INTER_CUBIC):
...
+        # # TODO: is this how we want to normalize output? Should we instead enforce model output to be normalized?
+        # NOTE: this normalization gets applied in the stacking function
+        # normalized_heatmaps = predict_heatmaps / predict_max
+        # normalized_heatmaps = normalized_heatmaps - predict_min
+        # normalized_heatmaps = normalized_heatmaps / np.max(normalized_heatmaps)
+        # return normalized_heatmaps
```
robertklee: sounds good

robertklee (approved):
LGTM, just added a comment about removing 2 print statements

evaluation.py
```diff
@@ -1,20 +1,38 @@
...
+        match = re.match(r'(.*)(_resume_.*$)', model_sub_dir)
+        if match:
+            self.output_sub_dir = os.path.join(output_base_dir, match.group(1), str(self.epoch))
+        else:
+            self.output_sub_dir = os.path.join(output_base_dir, self.model_sub_dir, str(self.epoch))
+        if not os.path.exists(self.output_sub_dir):
+            os.makedirs(self.output_sub_dir)
+
+        self.model_json, self.weights = util.find_resume_json_weights_str(model_base_dir, model_sub_dir, epoch)
+        self.num_hg_blocks = int(re.match(r'.*stacks_(\d\d)_.*',self.model_json).group(1))
```
robertklee: handle arbitrary number 
```suggestion
        self.num_hg_blocks = int(re.match(r'.*stacks_([\d]+)_.*',self.model_json).group(1))
```

util.py
```diff
@@ -5,25 +5,31 @@
...
-    args.resume_json = os.path.join(enclosing_dir, model_jsons[0])
-    args.resume_weights = os.path.join(enclosing_dir, model_saved_weights[0])
+    resume_json = os.path.join(enclosing_dir, model_jsons[0])
+    resume_weights = os.path.join(enclosing_dir, model_saved_weights[0])
+
+    print('Found model json: {}'.format(resume_json))
+    print('Found model weights for epoch {epoch:02d}: {weight_file_name}'.format(epoch=resume_epoch, weight_file_name=resume_weights))
+
+    assert os.path.exists(resume_json)
+    assert os.path.exists(resume_weights)
```
robertklee, line 29: just as a FYI, if my PR merges before yours, just `git checkout --ours util.py` when you rebase on main (take main's `util.py`). I made some changes that conflict with this, but won't affect any of your branch otherwise.

evaluation.py
```diff
@@ -1,20 +1,38 @@
...
+        self.epoch=epoch
+        match = re.match(r'(.*)(_resume_.*$)', model_sub_dir)
+        if match:
+            self.output_sub_dir = os.path.join(output_base_dir, match.group(1), str(self.epoch))
+        else:
+            self.output_sub_dir = os.path.join(output_base_dir, self.model_sub_dir, str(self.epoch))
+        if not os.path.exists(self.output_sub_dir):
+            os.makedirs(self.output_sub_dir)
+
+        self.model_json, self.weights = util.find_resume_json_weights_str(model_base_dir, model_sub_dir, epoch)
```
robertklee, line 30: if mine goes in first, you'll need to make this change:
```suggestion
        self.model_json, self.weights, _ = util.find_resume_json_weights_str(model_base_dir, model_sub_dir, epoch)
```

PR #17, opened 2021-03-26
QOL improvements: add helper function to load arbitrary img, update `notes` flag in training script

robertklee:
This is a very small PR that adds some additional functionality to the eval code to load arbitrary images.

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -1,9 +1,11 @@
-from HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
-from constants import *
-
+import cv2
 import matplotlib.pyplot as plt
 import numpy as np
-import cv2
+from PIL import Image
+
+from constants import *
+from HeatMap import HeatMap  # https://github.com/LinShanify/HeatMap
+
 
 class Evaluation():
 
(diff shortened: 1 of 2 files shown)


nicolepeverley (approved):
LGTM 👍

robertklee:
lol i forgot to push the important actual commit 😂 my bad

evaluation.py
```diff
@@ -35,7 +37,18 @@ def predict_heatmaps(self, h, X):
...
+        predict_mean = np.mean(predict_heatmaps)
+        predict_max = np.max(predict_heatmaps)
+        predict_min = np.min(predict_heatmaps)
+        predict_var = np.var(predict_heatmaps.flatten())
+        print('Mean: {:0.6e}\t Max: {:e}\t Min: {:e}\t Variance: {:e}'.format(predict_mean, predict_max, predict_min, predict_var))
+        normalized_heatmaps = predict_heatmaps / predict_max
+        normalized_heatmaps = normalized_heatmaps - predict_min
+        normalized_heatmaps = predict_heatmaps / np.max(predict_heatmaps)
+
+        return normalized_heatmaps
```
nicolepeverley, line 51: Do the normalized predicted heatmaps look different or is this just to normalize the np array?

nicolepeverley (commented):
lol I thought you were just updating the comments haha, still looks good just had a  question!

evaluation.py
robertklee: this is to normalize the array. We're getting some major issues with large numbers in the network right now so this is to try to combat that in order to plot it.

robertklee:
haha i had soft reset it to amend my commit, but forgot to force push afterwards

evaluation.py
robertklee: do you remember the really odd looking heat map? So normally our network output is ~0 to ~1, but occasionally it outputs like 6.0e16 and -6.0e16 (which results in that really blurry image we saw)

nicolepeverley: oh right sweet so this should keep it between the normal output

robertklee: yep exactly. I have to fix our model too so we dont even get to this state

robertklee:
gonna merge since it's a small change. lmk if i missed anything and I'll add it to my next PR!

evaluation.py
```diff
@@ -35,7 +37,18 @@ def predict_heatmaps(self, h, X):
...
-        return predict_heatmaps
+
+        print('model prediction metrics: ')
+        predict_mean = np.mean(predict_heatmaps)
+        predict_max = np.max(predict_heatmaps)
+        predict_min = np.min(predict_heatmaps)
+        predict_var = np.var(predict_heatmaps.flatten())
+        print('Mean: {:0.6e}\t Max: {:e}\t Min: {:e}\t Variance: {:e}'.format(predict_mean, predict_max, predict_min, predict_var))
+        normalized_heatmaps = predict_heatmaps / predict_max
+        normalized_heatmaps = normalized_heatmaps - predict_min
```
julianrocha, line 48: These two lines get overwritten by the next line

PR #19, opened 2021-03-27
Add Heatmap to Keypoint Functionality

nicolepeverley:
This PR adds the functionality for getting predicted keypoints from our predicted heatmaps output.

To test I have plotted the predicted keypoints onto the bounding box image.

@Airborne95 will follow up with a PR to transfer these keypoints back to the original image's scale.

Props to "test lady Janine"
img

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -75,3 +75,6 @@ VAL_SHUFFLE = False
 
 # Data filtering constants
 BBOX_MIN_SIZE = 900 # Filter out images smaller than 30x30, TODO tweak
+
+# Output filtering constants
+HM_TO_KP_THRESHOLD = 0.05
--- a/constants.py
+++ b/constants.py
@@ -67,6 +67,7 @@ BBOX_SLACK = 1.3 # before augmentation, increase bbox size to 130%
 
 # NOTE: The effective sigma is downscaled by a factor of 4 (from (256,256) to (64,64)) on each side, so ensure the sigma is appropriately sized
 HEATMAP_SIGMA = 4 # As per https://towardsdatascience.com/human-pose-estimation-with-stacked-hourglass-network-and-tensorflow-c4e9f84fd3ce
+REVERSE_HEATMAP_SIGMA = 1 # Use reverse sigma when heatmap is reversed by a factor of 4 from (64,64) to (256,256)
 
 # NOTE: Don't use, the scale is applied in the loss function.
 HEATMAP_SCALE = 1 #TODO figure out if we want to scale heatmaps, set to 1 for no effect. There are 82 times background pixels to foreground pixels in 7*7 patch of 64*64 heatmap, see same link for HEATMAP_SIGMA
@@ -77,4 +78,4 @@ VAL_SHUFFLE = False
 BBOX_MIN_SIZE = 900 # Filter out images smaller than 30x30, TODO tweak
 
 # Output filtering constants
-HM_TO_KP_THRESHOLD = 0.05
+HM_TO_KP_THRESHOLD = 0.2
--- a/evaluation.py
+++ b/evaluation.py
@@ -1,4 +1,11 @@
+<<<<<<< HEAD
 import cv2
+=======
+from HeatMap import HeatMap # https://github.com/LinShanify/HeatMap
+from constants import *
+from scipy.ndimage import gaussian_filter, maximum_filter
+
+>>>>>>> Added heatmap to keypoint functionality
 import matplotlib.pyplot as plt
 import numpy as np
 from PIL import Image
@@ -135,3 +142,38 @@ def load_and_preprocess_img(img_path, num_hg_blocks, x=None, y=None, w=None, h=N
     y_batch = [np.zeros((1, *(OUTPUT_DIM), NUM_COCO_KEYPOINTS)) for _ in range(num_hg_blocks)]
 
     return X_batch, y_batch
+
+    # Resources for heatmaps to keypoints
+    # https://github.com/yuanyuanli85/Stacked_Hourglass_Network_Keras/blob/eddf0ae15715a88d7859847cfff5f5092b260ae1/src/eval/heatmap_process.py#L5
+    # https://github.com/david8862/tf-keras-stacked-hourglass-keypoint-detection/blob/56707252501c73b2bf2aac8fff3e22760fd47dca/hourglass/postprocess.py#L17
+   
+    ### Returns np array of predicted keypoints from one image's heatmaps
+    def heatmaps_to_keypoints(self, heatmaps, threshold=1e-2):
+        keypoints = list()
+        for i in range(NUM_COCO_KEYPOINTS):
+            hmap = heatmaps[:,:,i]
+            # Do a heatmap blur with gaussian_filter
+            hmap = gaussian_filter(hmap, HEATMAP_SIGMA)
+
+            # Resize heatmap from Output DIM to Input DIM
+            resized_hmap = cv2.resize(hmap, INPUT_DIM, interpolation = cv2.INTER_AREA)
+
+            # Get peak point (brightest area) in heatmap with 3x3 max filter
+            peaks = self._non_max_supression(resized_hmap, windowSize=3, threshold=1e-2)
+
+            # Choose the max point in heatmap (we only pick 1 keypoint in each heatmap)
+            # and get its coordinates and confidence
+            y, x = np.where(peaks == peaks.max())
+            if len(x) > 0 and len(y) > 0:
+                keypoints.append((int(x[0]), int(y[0]), peaks[y[0], x[0]]))
+            else:
+                keypoints.append((0, 0, 0))
+        # Turn keypoints into np array
+        keypoints = np.array(keypoints)
+        return keypoints
+
+    def _non_max_supression(self, plain, windowSize=3, threshold=1e-2):
+        # Clear values less than threshold
+        under_thresh_indices = plain < threshold
+        plain[under_thresh_indices] = 0
+        return plain * (plain == maximum_filter(plain, footprint=np.ones((windowSize, windowSize))))
--- a/evaluation.py
+++ b/evaluation.py
@@ -99,7 +99,42 @@ class Evaluation():
         img_v_resize = self._vstack_images(heatmap_imgs) 
         
         cv2.imwrite(filename, img_v_resize) 
-    
+
+    # Resources for heatmaps to keypoints
+    # https://github.com/yuanyuanli85/Stacked_Hourglass_Network_Keras/blob/eddf0ae15715a88d7859847cfff5f5092b260ae1/src/eval/heatmap_process.py#L5
+    # https://github.com/david8862/tf-keras-stacked-hourglass-keypoint-detection/blob/56707252501c73b2bf2aac8fff3e22760fd47dca/hourglass/postprocess.py#L17
+   
+    ### Returns np array of predicted keypoints from one image's heatmaps
+    def heatmaps_to_keypoints(self, heatmaps, threshold=1e-2):
+        keypoints = list()
+        for i in range(NUM_COCO_KEYPOINTS):
+            hmap = heatmaps[:,:,i]
+            
+            # Do a heatmap blur with gaussian_filter
+            hmap = gaussian_filter(hmap, HEATMAP_SIGMA)
+             # Resize heatmap from Output DIM to Input DIM
+            resized_hmap = cv2.resize(hmap, INPUT_DIM, interpolation = cv2.INTER_AREA)
+
+            # Get peak point (brightest area) in heatmap with 3x3 max filter
+            peaks = self._non_max_supression(resized_hmap, windowSize=3, threshold=1e-2)
+
+            # Choose the max point in heatmap (we only pick 1 keypoint in each heatmap)
+            # and get its coordinates and confidence
+            y, x = np.where(peaks == peaks.max())
+            if len(x) > 0 and len(y) > 0:
+                keypoints.append((int(x[0]), int(y[0]), peaks[y[0], x[0]]))
+            else:
+                keypoints.append((0, 0, 0))
+        # Turn keypoints into np array
+        keypoints = np.array(keypoints)
+        print(keypoints)
+        return keypoints
+
+    def _non_max_supression(self, plain, windowSize=3, threshold=1e-2):
+        # Clear values less than threshold
+        under_thresh_indices = plain < threshold
+        plain[under_thresh_indices] = 0
+        return plain * (plain == maximum_filter(plain, footprint=np.ones((windowSize, windowSize))))
 
 """
 Runs the model for any general file. This aims to extend the DataGenerator output format for arbitrary images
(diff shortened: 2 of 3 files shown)


evaluation.py
```diff
@@ -135,3 +137,38 @@ def load_and_preprocess_img(img_path, num_hg_blocks, x=None, y=None, w=None, h=N
...
+    ### Returns np array of predicted keypoints from one image's heatmaps
+    def heatmaps_to_keypoints(self, heatmaps, threshold=1e-2):
+        keypoints = list()
+        for i in range(NUM_COCO_KEYPOINTS):
+            hmap = heatmaps[:,:,i]
+            # Do a heatmap blur with gaussian_filter
+            hmap = gaussian_filter(hmap, HEATMAP_SIGMA)
+
+            # Resize heatmap from Output DIM to Input DIM
+            resized_hmap = cv2.resize(hmap, INPUT_DIM, interpolation = cv2.INTER_AREA)
```
robertklee: consider resizing first before blur. we retain more info

evaluation.py
```diff
@@ -135,3 +137,38 @@ def load_and_preprocess_img(img_path, num_hg_blocks, x=None, y=None, w=None, h=N
...
+
+            # Get peak point (brightest area) in heatmap with 3x3 max filter
+            peaks = self._non_max_supression(resized_hmap, windowSize=3, threshold=1e-2)
+
+            # Choose the max point in heatmap (we only pick 1 keypoint in each heatmap)
+            # and get its coordinates and confidence
+            y, x = np.where(peaks == peaks.max())
+            if len(x) > 0 and len(y) > 0:
+                keypoints.append((int(x[0]), int(y[0]), peaks[y[0], x[0]]))
```
robertklee: what is the additional param here? a confidence level of sorts?

evaluation.py
```diff
@@ -135,3 +137,38 @@ def load_and_preprocess_img(img_path, num_hg_blocks, x=None, y=None, w=None, h=N
...
 
     return X_batch, y_batch
+
+    # Resources for heatmaps to keypoints
+    # https://github.com/yuanyuanli85/Stacked_Hourglass_Network_Keras/blob/eddf0ae15715a88d7859847cfff5f5092b260ae1/src/eval/heatmap_process.py#L5
+    # https://github.com/david8862/tf-keras-stacked-hourglass-keypoint-detection/blob/56707252501c73b2bf2aac8fff3e22760fd47dca/hourglass/postprocess.py#L17
+   
+    ### Returns np array of predicted keypoints from one image's heatmaps
+    def heatmaps_to_keypoints(self, heatmaps, threshold=1e-2):
+        keypoints = list()
```
robertklee: would recommend using a `np.array` here since we use it heavily in other sections of the code

run_evaluation.py
```diff
@@ -48,4 +48,45 @@
...
+# Clear plot image
+plt.clf()
+# Plot predicted keypoints on bounding box image
+x = []
+y = []
+for i in range(NUM_COCO_KEYPOINTS):
+    if(keypoints[i,0] != 0 and keypoints[i,1] != 0):
+      x.append(keypoints[i,0])
+      y.append(keypoints[i,1])
+plt.plot(x,y)
```
robertklee: I'd do a scatter plot tbh

robertklee: Until we get the right linking order done

robertklee (commented):
nice! left a few comments

evaluation.py
nicolepeverley: yeah so this I took from the other repo it's a confidence level, not sure if we want to use this for any reason or just append a 1 or 2 as the v number? (this doesn't get used by the OKS evaluation)

evaluation.py
nicolepeverley: `keypoints` gets converted back to an `np.array` at the end of the function!

evaluation.py
robertklee: yeah let's keep it for now. we can remove the kp later if it's not confident enough

evaluation.py
```diff
@@ -97,7 +99,40 @@ def save_stacked_evaluation_heatmaps(self, h, X, y, stacked_predict_heatmaps_fil
...
         
         cv2.imwrite(filename, img_v_resize) 
-    
+
+    # Resources for heatmaps to keypoints
+    # https://github.com/yuanyuanli85/Stacked_Hourglass_Network_Keras/blob/eddf0ae15715a88d7859847cfff5f5092b260ae1/src/eval/heatmap_process.py#L5
+    # https://github.com/david8862/tf-keras-stacked-hourglass-keypoint-detection/blob/56707252501c73b2bf2aac8fff3e22760fd47dca/hourglass/postprocess.py#L17
+   
+    ### Returns np array of predicted keypoints from one image's heatmaps
+    def heatmaps_to_keypoints(self, heatmaps, threshold=0.05):
```
nicolepeverley: I tested a couple images but we might need to test some more heatmap output to see what the best threshold is here!

robertklee: Sounds good! Want to throw that in `constants.py`?

evaluation.py
```diff
@@ -97,7 +99,40 @@ def save_stacked_evaluation_heatmaps(self, h, X, y, stacked_predict_heatmaps_fil
...
+        keypoints = list()
+        for i in range(NUM_COCO_KEYPOINTS):
+            hmap = heatmaps[:,:,i]
+            # Resize heatmap from Output DIM to Input DIM
+            resized_hmap = cv2.resize(hmap, INPUT_DIM, interpolation = cv2.INTER_AREA)
+            # Do a heatmap blur with gaussian_filter
+            resized_hmap = gaussian_filter(resized_hmap, HEATMAP_SIGMA)
+
+            # Get peak point (brightest area) in heatmap with 3x3 max filter
+            peaks = self._non_max_supression(resized_hmap, windowSize=3, threshold=HM_TO_KP_THRESHOLD)
```
robertklee: I think you might have missed using the parameter that's been passed in 
```suggestion
            peaks = self._non_max_supression(resized_hmap, windowSize=3, threshold=threshold)
```

robertklee (approved):
One small bug I think. Otherwise lgtm

evaluation.py
```diff
@@ -97,7 +98,40 @@ def save_stacked_evaluation_heatmaps(self, h, X, y, stacked_predict_heatmaps_fil
...
+    # https://github.com/yuanyuanli85/Stacked_Hourglass_Network_Keras/blob/eddf0ae15715a88d7859847cfff5f5092b260ae1/src/eval/heatmap_process.py#L5
+    # https://github.com/david8862/tf-keras-stacked-hourglass-keypoint-detection/blob/56707252501c73b2bf2aac8fff3e22760fd47dca/hourglass/postprocess.py#L17
+   
+    ### Returns np array of predicted keypoints from one image's heatmaps
+    def heatmaps_to_keypoints(self, heatmaps, threshold=HM_TO_KP_THRESHOLD):
+        keypoints = list()
+        for i in range(NUM_COCO_KEYPOINTS):
+            hmap = heatmaps[:,:,i]
+            # Resize heatmap from Output DIM to Input DIM
+            resized_hmap = cv2.resize(hmap, INPUT_DIM, interpolation = cv2.INTER_AREA)
```
julianrocha: According to [this](https://stackoverflow.com/questions/23853632/which-kind-of-interpolation-best-for-resizing-image), when enlarging an image, we should use `INTER_LINEAR` (computationally faster but less quality) OR `INTER_CUBIC` (computationally slower but higher quality). `INTER_LINEAR` is default for cv2 so we may wanna just roll with this. I've made a note to go back to the data generator and apply `INTER_AREA` which is better for shrinking images.

evaluation.py
```diff
@@ -97,7 +98,40 @@ def save_stacked_evaluation_heatmaps(self, h, X, y, stacked_predict_heatmaps_fil
...
+            # Do a heatmap blur with gaussian_filter
+            resized_hmap = gaussian_filter(resized_hmap, HEATMAP_SIGMA)
```
julianrocha: How is this sigma chosen. We have `HEATMAP_SIGMA` = 4 because it gets applied before shrinking the image at data generator. Since we likely wanna blur with the intention of achieving 7x7, we might wanna pick sigma = 1 since this image is at full resolution.

evaluation.py
```diff
@@ -97,7 +98,40 @@ def save_stacked_evaluation_heatmaps(self, h, X, y, stacked_predict_heatmaps_fil
...
+            # Get peak point (brightest area) in heatmap with 3x3 max filter
+            peaks = self._non_max_supression(resized_hmap, threshold, windowSize=3)
+
+            # Choose the max point in heatmap (we only pick 1 keypoint in each heatmap)
+            # and get its coordinates and confidence
+            y, x = np.where(peaks == peaks.max())
+            if len(x) > 0 and len(y) > 0:
+                keypoints.append((int(x[0]), int(y[0]), peaks[y[0], x[0]]))
+            else:
+                keypoints.append((0, 0, 0))
```
julianrocha, line 125: I may be mistaken, but won't `np.where(peaks == peaks.max())` always return and x, y (since every array has a max value)? Has this been tested to mark a weak heatmap as an unlabelled kp?

constants.py
```diff
@@ -75,3 +75,6 @@ class ImageAugmentationStrength(Enum):
 
 # Data filtering constants
 BBOX_MIN_SIZE = 900 # Filter out images smaller than 30x30, TODO tweak
+
+# Output filtering constants
+HM_TO_KP_THRESHOLD = 0.05
```
julianrocha: Likely a reasonable threshold, hopefully the visualization tests paired with the kp skeletons can give us insight into tuning this.

run_evaluation.py
```diff
@@ -48,4 +48,45 @@
...
+hm = HeatMap(X,heatmap)
+img = np.array(hm.image)
+
+# Clear plot image
+plt.clf()
+# Plot predicted keypoints on bounding box image
+x = []
+y = []
+for i in range(NUM_COCO_KEYPOINTS):
+    if(keypoints[i,0] != 0 and keypoints[i,1] != 0):
```
julianrocha, line 87: We should be using the third `v` flag to determine if a kp is labelled or not.

run_evaluation.py
```diff
@@ -48,4 +48,45 @@
...
+            batch_size=1,
+            online_fetch=False)
+
+# Select image to predict heatmaps
+X_batch, y_stacked = generator[168] # choose one image for evaluation
+y_batch = y_stacked[0] # take first hourglass section
+X, y = X_batch[0], y_batch[0] # take first example of batch
+
+# Get predicted heatmaps for image
+predict_heatmaps=eval.predict_heatmaps(h, X)
```
julianrocha, line 71: This may be the one area that conflicts with my branch, predict_heatmaps will take in a batch instead of a single example. I think it would be easiest if we merge your branch first and I fix this line in my branch after rebasing.

julianrocha (approved):
Left a few comments related to tuning and labelled vs unlabelled kps. Looks great otherwise, glad we could find some existing repos for guidance. The actual conversion function ended up being much simpler than I thought it would. 👍  P.S I think we should merge this before mine

evaluation.py
nicolepeverley: The unlabelled kp's are turned to 0 under the threshold in `_non_max_supression` which is why we may have to test out the threshold as we go. It will label 0,0,0 for any heatmap under the given threshold.

evaluation.py
nicolepeverley: This is a good point I forgot about reversing the size for sigma I'll change it to 1 👍

evaluation.py
nicolepeverley: Good catch!

constants.py
robertklee: I suspect a higher threshold is needed. We don't want uncertain activations, right?

nicolepeverley: I'm just doing some testing and will likely raise it until it drops active keypoints!

robertklee: solid approach :)

evaluation.py
robertklee: good catch

julianrocha (approved):
noice

evaluation.py
```diff
@@ -97,7 +98,40 @@ def save_stacked_evaluation_heatmaps(self, h, X, y, stacked_predict_heatmaps_fil
...
+
+            # Choose the max point in heatmap (we only pick 1 keypoint in each heatmap)
+            # and get its coordinates and confidence
+            y, x = np.where(peaks == peaks.max())
+            if int(x[0]) > 0 and int(y[0]) > 0:
+                keypoints.append((int(x[0]), int(y[0]), 1))
+            else:
+                keypoints.append((0, 0, 0))
+        # Turn keypoints into np array
+        keypoints = np.array(keypoints)
```
robertklee, line 127: Are you intending for the shape to be nd.array of shape (point, xyz)?

robertklee: I just saw this now. I think `keypoints.append(tuple)` is mixing tuples and lists. Is this intentional?

robertklee:
Awesome! I think we'll need to figure out what to do with two activations in the same output (when the model confuses left and right). It happens pretty frequently unfortunately. That'll probably involve some extra logic. Another potential thing is doing centre of mass (taking the centre to be a weighted sum of the circle of activations). 

@julianrocha @nicolepeverley thoughts?

evaluation.py
nicolepeverley: I think it should add tuples to a list and then convert to an nd array of points [x,y,z] do you think that should be handled a different way?

nicolepeverley:
> Awesome! I think we'll need to figure out what to do with two activations in the same output (when the model confuses left and right). It happens pretty frequently unfortunately. That'll probably involve some extra logic. Another potential thing is doing centre of mass (taking the centre to be a weighted sum of the circle of activations).
> 
> @julianrocha @nicolepeverley thoughts?

Yeah it'll be interesting to see if it picks the brightest spot, whether it mixes up the left and right, maybe we should keep an eye on how many times this is happening and possibly adjust the threshold or decision for brightest point it makes!

evaluation.py
robertklee: I just ran a little test in my python terminal and the end result is [ [x1, y1, v1], [x2, y2, v2], etc] (so it's a 2d array). I just wanted to check because the original COCO data was a 1d array of `[x1,y1,v1,x2,y2,v2,...]`.

nicolepeverley: Oh yeah it still has to get adjusted with Rafay's undoing bounding box code which formats it like that I believe 👍 we can change it here to but it just depends on how it gets fed into that part

robertklee: Sounds good. no preference from my side

nicolepeverley: Sounds good I'll mention it to Rafay thanks for checking!

julianrocha:
> Awesome! I think we'll need to figure out what to do with two activations in the same output (when the model confuses left and right). It happens pretty frequently unfortunately. That'll probably involve some extra logic. Another potential thing is doing centre of mass (taking the centre to be a weighted sum of the circle of activations).

It appears as though the R/L issues we're having are producing almost identical heatmaps for R/L, with no cloud with clearly higher activation in the R heatmap and not the L. This is based on the visualizations I've been looking at. I do feel like adding logic to the heatmap->kp conversion for L to pick the opposite cloud that R picked is a bit of a bandaid over the underlying problem. which is the activations should be higher for the proper side, and the model should have some concept of R vs. L. In terms of center of mass logic, this might be necessary for the clouds that are not very confident.

robertklee:
I think the lowest hanging fruit is to ensure we're actually feeding the right shape heat maps. Next is probably trying the normal MSE or a lower weighted MSE (instead of 82, maybe try 41). I suspect the model is learning that if it guesses 0 and gets it wrong, its a worse penalty than guessing both at the same time and guaranteeing that it gets one of them right.

julianrocha:
> I think the lowest hanging fruit is to ensure we're actually feeding the right shape heat maps. Next is probably trying the normal MSE or a lower weighted MSE (instead of 82, maybe try 41). I suspect the model is learning that if it guesses 0 and gets it wrong, its a worse penalty than guessing both at the same time and guaranteeing that it gets one of them right.

There are so many ways this model can find local minima it's crazy. I think this is a really sound hypothesis though, we should definitely look into visualizing a series of WMSE models with different W. Maybe train on subset 0.3-0.5 and visualize some of the earlier epochs. We could also potentially extend the representative set to include more difficult R/L situations.

PR #20, opened 2021-03-28
Add a number of model training enhancements: optimizer, activation func, learning rate, and kp filter

robertklee:
This is to update our model to attempt to resolve two problems: extreme unbounded predictions, slow training times

The main additions are:
* Optional change to activation function to use sigmoid instead of linear for the prediction layer. This has the benefit of bounding our output to [0,1], but has the drawback of potentially vanishing gradients for large x values (which has a gradient of 0).
* Change in our optimizer between RMSProp and Adam. Right now we use RMSProp, but Adam may be faster. 
* Increase learning rate by a factor of 10. We can add a scheduler to reduce it later on, but we are training way too slow. The pytorch repo uses 1e-2 where we use 5e-3. It used to be 5e-4
* Filter out annotations <= 4 keypoints. These are mostly garbage data and could adversely affect our model. We can still generalize pretty well since there's still a lot of high quality annotations >=5.

## Quality of life and dev experience

* adds a pretty big refactoring of hourglass.py because the parameters were getting out of hand. Everything is moved to the constructor now.
* makes our subset deterministic by seeding it with a state.
* automatically find highest epoch number if it is not provided
* change resume logic to use regex instead of hardcoded pattern

This knocks off 3 of our backlog items


evaluation.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, model_sub_dir, epoch, model_base_dir=DEFAULT_MODEL_BASE_DIR,
         if not os.path.exists(self.output_sub_dir):
             os.makedirs(self.output_sub_dir)
 
-        self.model_json, self.weights = util.find_resume_json_weights_str(model_base_dir, model_sub_dir, epoch)
+        self.model_json, self.weights, _ = util.find_resume_json_weights_str(model_base_dir, model_sub_dir, epoch)
```
Airborne95, line 30: What does this do?

robertklee: `util.find_resume_json_weights_str` returns a tuple of length 3 now so the `_` is required to unpack it properly

colab_train.ipynb
```diff
@@ -220,17 +220,23 @@
         "outputId": "1f977904-3e1f-4e7d-c0ac-3d8511514c14"
       },
       "source": [
-        "!python3 train.py --epochs 100 --batch 12 --hourglass 4 --subset 1.0 --loss \"weighted_mse\" --augment \"none\"\n",
+        "!python3 train.py --epochs 100 --batch 16 --hourglass 4 --subset 1.0 --loss \"weighted_mse\" --augment \"light\"\n",
         "\n",
         "# To resume, add the following to the above command, and point it to the subdirectory of the model checkpoints. Ensure everything matches\n",
         "# --resume True --resume-epoch 42 --resume-subdir \"2021-03-14-15h-92m_batchsize_12_hg_4\"\n",
         "\n",
+        "# The options are below. NOTE that ALL options must match across training sessions:\n",
```
nicolepeverley, line 228: This is super helpful 💯

hourglass.py
```diff
@@ -46,26 +56,33 @@ def load_and_filter_annotations(self, path_to_train_anns,path_to_val_anns, subse
...
-            train_df = train_df.sample(frac=subset)
+            train_df = train_df.sample(frac=subset, random_state=1)
         print(f"Train/Val dfs contains {len(train_df)}/{len(val_df)} anns")
         return train_df.reset_index(), val_df.reset_index()
 
-    def _start_train(self, batch_size, model_base_dir, epochs, initial_epoch, model_subdir, current_time, subset, loss_str, image_aug_str, pickle_name):
-        loss = get_loss_from_string(loss_str)
-        self._compile_model(loss)
+    def _start_train(self, batch_size, model_base_dir, epochs, initial_epoch, model_subdir, current_time, subset):
+        print(f'Loss function selected:               {self.loss_str}')
```
nicolepeverley, line 69: Good idea to print the parameters!

util.py
```diff
@@ -4,32 +4,63 @@
...
+        print('OutputActivation was not found in possible options.')
+        print('Available options are:')
+        available_activs = [name for name, member in OutputActivation.__members__.items()]
+        print(available_activs)
+        exit(1)
+    return activ
 
 def find_resume_json_weights_args(args):
-    return find_resume_json_weights_str(args.model_save, args.resume_subdir, args.resume_epoch)
+    args.resume_json, args.resume_weights, args.resume_epoch = find_resume_json_weights_str(args.model_save, args.resume_subdir, args.resume_epoch)
```
Airborne95, line 30: Just double checking, does this no longer need to be returned?

nicolepeverley (approved):
Nice clean up LGTM 👍

Airborne95 (approved):

util.py
```diff
@@ -26,25 +26,41 @@ def validate_activation(activation):
         exit(1)
     return activ
 
-def find_resume_json_weights(args):
-    enclosing_dir = os.path.join(args.model_save, args.resume_subdir)
+def find_resume_json_weights_args(args):
+    args.model_save, args.resume_subdir, args.resume_epoch = find_resume_json_weights_str(args.model_save, args.resume_subdir, args.resume_epoch)
+
+def find_resume_json_weights_str(model_base_dir, model_subdir, resume_epoch):
```
Ckoelewyn, line 32: I like this

Ckoelewyn (approved):
Changes look good.

util.py
robertklee: Yeah it was a bug. Shouldn't have been returned in the first place

PR #21, opened 2021-03-30
Very small PR to make it easier to copy the training session ID to resume 

robertklee:
This is a tiny change to make it easier to find the subdir to copy and paste for resuming training.


julianrocha (approved):
very risky but lgtm

Ckoelewyn (approved):
✨

PR #23, opened 2021-03-31
Enable reusing the same subdirectory for model resumes **only if** it is resuming the latest epoch file

robertklee:
This adds logic to reuse the same training session ID on resuming training, under the precondition that the model is resuming the newest epoch weight file. This is to avoid weird problems with Tensorboard logs when resuming from an earlier epoch. 

This also refactors all `.format()` in `hourglass.py` to use `f` formatting.

### DIFF
--- a/train.py
+++ b/train.py
@@ -99,6 +99,9 @@ def process_args():
     argparser.add_argument('--resume-subdir',
                         default=None,
                         help='Subdirectory containing architecture json and weights')
+    argparser.add_argument('--resume-with-new-run',
+                        default=False,
+                        help='start a new session ID on resume. Default will be true if resume epoch is not the latest weight file.')
     argparser.add_argument('--pickle',
                         default=None,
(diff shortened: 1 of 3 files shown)


train.py
```diff
@@ -116,7 +119,14 @@ def process_args():
     if args.resume:
         # Automatically locate architecture json and model weights
         if args.resume_subdir is not None:
-            find_resume_json_weights_args(args)
+            args.resume_json, args.resume_weights, args.resume_epoch = find_resume_json_weights_str(args.model_save, args.resume_subdir, args.resume_epoch)
+
+        # If we are not resuming from the highest epoch in that subdir, start a new run
```
nicolepeverley, line 124: Sounds good 👍

nicolepeverley (approved):

robertklee:
I've just tested the logic locally to make sure everything works. :)

julianrocha (approved):
lgtm, thx for testing

PR #24, opened 2021-03-31
Undo bounding box

Airborne95:
This PR introduces the following functionality: 
- Undo bounding box transformations on keypoints 
- Create a json file for COCO Eval 
- Run OKS Summary 

**Note:**  to use COCO eval we need numpy 1.17.0. To run this code you need to run the following commands:
`pip3 install -r requirements.txt`
`pip3 uninstall pycocotools`
`pip3 install --upgrade pycocotools`

If the above does not work then try uninstalling numpy manually and reinstalling it with the following commands and then try the above again:
`pip3 uninstall numpy`
`pip3 install --upgrade numpy==1.17.0`

### DIFF
--- a/data_generator.py
+++ b/data_generator.py
@@ -236,6 +236,7 @@ class DataGenerator(Sequence):
             if(self.is_eval):
                 metadata = {}
                 metadata["src_set_image_id"] = ann['src_set_image_id']
+                metadata["ann_id"] = ann['ann_id']
                 metadata["coco_url"] = ann['coco_url']
                 metadata["cropped_width"] = cropped_width
                 metadata["cropped_height"] = cropped_height
--- a/data_generator.py
+++ b/data_generator.py
@@ -233,7 +233,7 @@ class DataGenerator(Sequence):
                 img, ann['bbox'])
             transformed_label = self.transform_label(
                 ann['keypoints'], cropped_width, cropped_height, anchor_x, anchor_y)
-            if(self.is_eval):
+            if self.is_eval:
                 metadata = {}
                 metadata["src_set_image_id"] = ann['src_set_image_id']
                 metadata["ann_id"] = ann['ann_id']
@@ -275,7 +275,7 @@ class DataGenerator(Sequence):
         for _ in range(self.num_hg_blocks):
             y_stacked.append(y)
 
-        if(self.is_eval):
+        if self.is_eval:
             return X, y_stacked, metadatas
 
         return X, y_stacked
--- a/data_generator.py
+++ b/data_generator.py
@@ -32,7 +32,7 @@ from data_augmentation import *
 class DataGenerator(Sequence):
 
     def __init__(self, df, base_dir, input_dim, output_dim, num_hg_blocks, shuffle=False, \
-        batch_size=DEFAULT_BATCH_SIZE, online_fetch=False, img_aug_strength=None):
+        batch_size=DEFAULT_BATCH_SIZE, online_fetch=False, img_aug_strength=None, is_eval=False):
 
         self.df = df                    # df of the the annotations we want
         self.base_dir = base_dir        # where to read imgs from in collab runtime
@@ -41,13 +41,14 @@ class DataGenerator(Sequence):
         self.num_hg_blocks = num_hg_blocks
         self.shuffle = shuffle
         self.batch_size = batch_size
+        self.is_eval = is_eval
         # If true, images will be loaded from url over network rather than filesystem
         self.online_fetch = online_fetch
         if img_aug_strength is not None:
-            self.augmenter = get_augmenter_pipeline(img_aug_strength) 
+            self.augmenter = get_augmenter_pipeline(img_aug_strength)
         else:
-            self.augmenter = None 
-        
+            self.augmenter = None
+
         self.on_epoch_end()
 
     # after each epoch, shuffle indices so data order changes
@@ -149,7 +150,7 @@ class DataGenerator(Sequence):
             else:
                 x = kp.x
                 y = kp.y
-                v = 1 
+                v = 1
 
             transformed_label.append(x)
             transformed_label.append(y)
@@ -215,6 +216,7 @@ class DataGenerator(Sequence):
         # Order of last dimension: (heatmap for each kp) repeated num_hg_blocks times
         y = np.empty((self.batch_size, *self.output_dim, NUM_COCO_KEYPOINTS))
 
+        metadatas = []
         # get the indices of the requested batch
         indices = self.indices[idx*self.batch_size:(idx+1)*self.batch_size]
         for i, data_index in enumerate(indices):
@@ -231,6 +233,19 @@ class DataGenerator(Sequence):
                 img, ann['bbox'])
             transformed_label = self.transform_label(
                 ann['keypoints'], cropped_width, cropped_height, anchor_x, anchor_y)
+            if(self.is_eval):
+                metadata = {}
+                metadata["src_set_image_id"] = ann['src_set_image_id']
+                metadata["coco_url"] = ann['coco_url']
+                metadata["cropped_width"] = cropped_width
+                metadata["cropped_height"] = cropped_height
+                metadata["anchor_x"] = anchor_x
+                metadata["anchor_y"] = anchor_y
+                metadata["input_dim"] = self.input_dim
+                metadata["output_dim"] = self.output_dim
+                metadata["transformed_label"] = transformed_label #DEBUG
+                metadata["ground_truth_keypoints"] = ann['keypoints'] #DEBUG
+                metadatas.append(metadata)
 
             # if image augmentations should be applied
             if self.augmenter is not None:
@@ -258,5 +273,8 @@ class DataGenerator(Sequence):
         y_stacked = []
         for _ in range(self.num_hg_blocks):
             y_stacked.append(y)
+
+        if(self.is_eval):
+            return X, y_stacked, metadatas
 
         return X, y_stacked
--- a/evaluation.py
+++ b/evaluation.py
@@ -187,6 +187,7 @@ class Evaluation():
         Example:  [1,2,0,1,4,666,32...]
     """
     def undo_bounding_box_transformations(self, metadata, untransformed_predictions):
+      untransformed_predictions = np.array(untransformed_predictions).flatten()
       predicted_labels = []
       for i in range(len(untransformed_predictions)):
         if( i % 3 == 0 ): # is an x-coord
@@ -203,6 +204,38 @@ class Evaluation():
       metadata['predicted_labels'] = predicted_labels
       return metadata
 
+    def write_to_json_file(self, list_of_predictions, location):
+        f = open(location, "w")
+        f.write('[')
+        for i in range(len(list_of_predictions)):
+            f.write(str(list_of_predictions[i]))
+            if(i < len(list_of_predictions) - 1):
+                f.write(',')
+        f.write(']')
+        f.close()
+
+    def create_oks_obj(self, metadata):
+        # TODO figure out score category
+        oks_obj = {}
+        oks_obj['image_id'] = metadata['src_set_image_id']
+        oks_obj['category_id'] = 1
+        oks_obj['keypoints'] = metadata['predicted_labels']
+        oks_obj['score'] = 0.502
+        return oks_obj
+
+    def predict_keypoints(self,generator, location):
+        list_of_predictions = []
+        for i in range(len(generator)):
+            j = 0
+            X_batch, y_stacked, metadatas = generator[i] # choose one image for evaluation
+            predict_heatmaps= self.predict_heatmaps(X_batch)
+            for X, metadata in zip(X_batch, metadatas):
+                keypoints = self.heatmaps_to_keypoints(predict_heatmaps[self.num_hg_blocks-1, j, :, :, :])
+                metadata = self.undo_bounding_box_transformations(metadata, keypoints)
+                list_of_predictions.append(self.create_oks_obj(metadata))
+                j+=1
+        self.write_to_json_file(list_of_predictions, location)
+        print('DONE')
 # ----------------------- End of Class -----------------------
 
 """
--- a/evaluation.py
+++ b/evaluation.py
@@ -103,12 +103,6 @@ class Evaluation():
         stacked_ground_truth_heatmaps = cv2.cvtColor(stacked_ground_truth_heatmaps, cv2.COLOR_BGRA2RGB)
         stacked_ground_truth_heatmaps = cv2.normalize(stacked_ground_truth_heatmaps, None, alpha=0, beta=255, norm_type=cv2.NORM_MINMAX, dtype=cv2.CV_8UC1)
 
-
-        # Save stacked images to disk
-        plt.imsave(stacked_predict_heatmaps_file, stacked_predict_heatmaps)
-        plt.imsave(stacked_ground_truth_heatmaps_file, stacked_ground_truth_heatmaps)
-        filename = filename
-
         heatmap_imgs = []
         heatmap_imgs.append(stacked_predict_heatmaps)
         heatmap_imgs.append(stacked_ground_truth_heatmaps)
--- a/evaluation.py
+++ b/evaluation.py
@@ -44,8 +44,9 @@ class Evaluation():
 
         # resizing images
         im_list_resize = [cv2.resize(img,
-                        (w_min, int(img.shape[0] * w_min / img.shape[1])),
-                                    interpolation = interpolation)
+                            (w_min,
+                            int(img.shape[0] * w_min / img.shape[1])),
+                            interpolation = interpolation)
                         for img in img_list]
         # return final image
         return cv2.vconcat(im_list_resize)
@@ -147,9 +148,6 @@ class Evaluation():
         under_thresh_indices = plain < threshold
         plain[under_thresh_indices] = 0
         return plain * (plain == maximum_filter(plain, footprint=np.ones((windowSize, windowSize))))
-        img_v_resize = self._vstack_images(heatmap_imgs)
-
-        cv2.imwrite(filename, img_v_resize)
 
     """
         Parameters
@@ -160,7 +158,7 @@ class Evaluation():
         untransformed_x : int
         x coordinate to
     """
-    def __undo_x(self, metadata, untransformed_x):
+    def _undo_x(self, metadata, untransformed_x):
       predicted_x = round(untransformed_x * metadata['cropped_width'] / metadata['input_dim'][0] + metadata['anchor_x'])
       return int(predicted_x)
 
@@ -173,7 +171,7 @@ class Evaluation():
         untransformed_y : int
         x coordinate to
     """
-    def __undo_y(self, metadata, untransformed_y):
+    def _undo_y(self, metadata, untransformed_y):
       predicted_y = round(untransformed_y * metadata['cropped_height'] / metadata['input_dim'][1] + metadata['anchor_y'])
       return int(predicted_y)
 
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -7,6 +7,6 @@ h5py==2.10.0
 opencv-contrib-python==4.4.0.42
 opencv-python==4.4.0.42
 matplotlib
-numpy
+numpy=1.17.0
```
robertklee: permit anything newer than 1.17
```suggestion
numpy>=1.17.0
```

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
         under_thresh_indices = plain < threshold
         plain[under_thresh_indices] = 0
         return plain * (plain == maximum_filter(plain, footprint=np.ones((windowSize, windowSize))))
+        img_v_resize = self._vstack_images(heatmap_imgs)
```
nicolepeverley: Is this supposed to be here after the return?

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+    def undo_bounding_box_transformations(self, metadata, untransformed_predictions):
+      untransformed_predictions = np.array(untransformed_predictions).flatten()
+      predicted_labels = []
+      list_of_scores = []
+      for i in range(len(untransformed_predictions)):
+        if( i % 3 == 0 ): # is an x-coord
+          predicted_labels.append(self.__undo_x(metadata, untransformed_predictions[i]))
+        elif( i % 3 == 1 ): # is a y-coord
+          predicted_labels.append(self.__undo_y(metadata, untransformed_predictions[i]))
+        elif( i % 3 == 2) : # is v (visbility)
```
nicolepeverley: Might want to update this comment to score!

run_evaluation.py
```diff
@@ -90,4 +90,77 @@
...
+            online_fetch=True,
+            is_eval=True)
+print("Created DataGen instance")
+
+image_ids = eval.predict_keypoints(generator, location='output/predictions.json')
+print('Doneee')
+
+
+# %%
+annFile = "person_keypoints_val2017.json"
```
nicolepeverley: Do we have this file saved anywhere in our data already or should we save it in the data folder?

robertklee: Good catch. We can use `DEFAULT_VAL_ANNOT_PATH` here. (This is automatically updated based on whether we train local or on cloud.

evaluation.py
```diff
@@ -134,7 +135,7 @@ def heatmaps_to_keypoints(self, heatmaps, threshold=HM_TO_KP_THRESHOLD):
             # and get its coordinates and confidence
             y, x = np.where(peaks == peaks.max())
             if int(x[0]) > 0 and int(y[0]) > 0:
```
julianrocha, line 137: Unrelated to your changes. Do we need to cast to int here? This would turn a 0.99 into 0. Is this what we want? @nicolepeverley

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+        oks_obj["keypoints"] = metadata['predicted_labels']
+        oks_obj["score"] = float(metadata['score']) # TODO figure out score category
+        return oks_obj
+
+    def predict_keypoints(self,generator, location):
+        list_of_predictions = []
+        image_ids = []
+        for i in range(len(generator)):
+            j = 0
+            X_batch, y_stacked, metadatas = generator[i] # choose one image for evaluation
```
julianrocha: You can do:
```suggestion
        for  X_batch, y_stacked, metadatas in generator:
            j = 0
```

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
         under_thresh_indices = plain < threshold
         plain[under_thresh_indices] = 0
         return plain * (plain == maximum_filter(plain, footprint=np.ones((windowSize, windowSize))))
+        img_v_resize = self._vstack_images(heatmap_imgs)
+
+        cv2.imwrite(filename, img_v_resize)
```
julianrocha: this 1 too

evaluation.py
nicolepeverley: I think I just left the int cast from the original code in the other repo (https://github.com/yuanyuanli85/Stacked_Hourglass_Network_Keras/blob/eddf0ae15715a88d7859847cfff5f5092b260ae1/src/eval/heatmap_process.py#L5). I'm pretty sure they should all already be ints. I just tested a couple images and they all appear to be 0 or another int within 255.

data_generator.py
```diff
@@ -231,6 +233,20 @@ def __getitem__(self, idx):
...
+                metadata["ann_id"] = ann['ann_id']
+                metadata["coco_url"] = ann['coco_url']
+                metadata["cropped_width"] = cropped_width
+                metadata["cropped_height"] = cropped_height
+                metadata["anchor_x"] = anchor_x
+                metadata["anchor_y"] = anchor_y
+                metadata["input_dim"] = self.input_dim
+                metadata["output_dim"] = self.output_dim
+                metadata["transformed_label"] = transformed_label #DEBUG
+                metadata["ground_truth_keypoints"] = ann['keypoints'] #DEBUG
```
robertklee, line 258: Use dictionary constructor so it's easier to read
```suggestion
                metadata = {
                    "src_set_image_id":         ann['src_set_image_id',
                    "ann_id":                   ann['ann_id'],
                    "coco_url":                 ann['coco_url'],
                    "cropped_width":            cropped_width,
                    "cropped_height":           cropped_height,
                    "anchor_x":                 anchor_x,
                    "anchor_y":                 anchor_y,
                    "input_dim":                self.input_dim,
                    "output_dim":               self.output_dim,
                    "transformed_label":        transformed_label, #DEBUG
                    "ground_truth_keypoints":   ann['keypoints'] #DEBUG
                }
```

data_generator.py
```diff
@@ -231,6 +233,20 @@ def __getitem__(self, idx):
                 img, ann['bbox'])
             transformed_label = self.transform_label(
                 ann['keypoints'], cropped_width, cropped_height, anchor_x, anchor_y)
+            if(self.is_eval):
```
robertklee: python code guidelines suggest no parentheses:
```suggestion
            if self.is_eval:
```

data_generator.py
```diff
@@ -259,4 +275,7 @@ def __getitem__(self, idx):
         for _ in range(self.num_hg_blocks):
             y_stacked.append(y)
 
+        if(self.is_eval):
```
robertklee:
```suggestion
        if self.is_eval:
```

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+    """
+        Parameters
+        ----------
+        metadata : object
+        should be metadata associated to a single image
+
+        untransformed_x : int
+        x coordinate to
+    """
+    def __undo_x(self, metadata, untransformed_x):
```
robertklee: Double `__` in python means name mangling (the interpreter automatically adds the class name in front). Is this intentional?

```suggestion
    def _undo_x(self, metadata, untransformed_x):
```

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+    """
+        Parameters
+        ----------
+        metadata : object
+        should be metadata associated to a single image
+
+        untransformed_y : int
+        x coordinate to
+    """
+    def __undo_y(self, metadata, untransformed_y):
```
robertklee: same here
```suggestion
    def _undo_y(self, metadata, untransformed_y):
```

run_evaluation.py
```diff
@@ -90,4 +90,77 @@
...
+cocoGt=COCO(annFile)
+
+resFile = "output/predictions.json"
```
robertklee, line 154: we'll need to save these in a unique path. (I'd recommend taking a look at Julian's PR #16  to see how he saves output files

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+        ----------
+        metadata : object
+        should be metadata associated to a single image
+
+        untransformed_predictions : list
+        a list of precitions that need to be transformed
+        Example:  [1,2,0,1,4,666,32...]
+    """
+    def undo_bounding_box_transformations(self, metadata, untransformed_predictions):
+      untransformed_predictions = np.array(untransformed_predictions).flatten()
```
robertklee: nit: we could just initialize it at first (in eval get keypoints) to be a 1d array

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+        elif( i % 3 == 1 ): # is a y-coord
+          predicted_labels.append(self.__undo_y(metadata, untransformed_predictions[i]))
+        elif( i % 3 == 2) : # is v (visbility)
+          if(untransformed_predictions[i] == 0):
+            predicted_labels[i-1] = 0 # Set y value to 0
+            predicted_labels[i-2] = 0 # Set x value to 0
+            predicted_labels.append(0)
+          else:
+            predicted_labels.append(1)
+            list_of_scores.append(untransformed_predictions[i])
```
robertklee: nit indenting to be consistent

evaluation.py
robertklee: nit: we could iterate from `i in range(NUM_COCO_KP)` here, and then get x,y,score from `i, i+1, i+2`. I think it would be a bit cleaner

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+      list_of_scores = []
+      for i in range(len(untransformed_predictions)):
+        if( i % 3 == 0 ): # is an x-coord
+          predicted_labels.append(self.__undo_x(metadata, untransformed_predictions[i]))
+        elif( i % 3 == 1 ): # is a y-coord
+          predicted_labels.append(self.__undo_y(metadata, untransformed_predictions[i]))
+        elif( i % 3 == 2) : # is v (visbility)
+          if(untransformed_predictions[i] == 0):
+            predicted_labels[i-1] = 0 # Set y value to 0
+            predicted_labels[i-2] = 0 # Set x value to 0
```
robertklee: then we wouldn't need to do this i-1 which is a bit dangerous. (If the transformed x is 0 for the first key point, we are indexing negative numbers)

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+        untransformed_predictions : list
+        a list of precitions that need to be transformed
+        Example:  [1,2,0,1,4,666,32...]
+    """
+    def undo_bounding_box_transformations(self, metadata, untransformed_predictions):
+      untransformed_predictions = np.array(untransformed_predictions).flatten()
+      predicted_labels = []
+      list_of_scores = []
+      for i in range(len(untransformed_predictions)):
+        if( i % 3 == 0 ): # is an x-coord
```
robertklee: nit: no brackets on if (same for a few places further down this file)

evaluation.py
```diff
@@ -38,16 +39,16 @@ def __init__(self, model_sub_dir, epoch, model_base_dir=DEFAULT_MODEL_BASE_DIR,
...
-        # return final image 
-        return cv2.vconcat(im_list_resize) 
+        # take minimum width
+        w_min = min(img.shape[1] for img in img_list)
+
+        # resizing images
+        im_list_resize = [cv2.resize(img,
+                        (w_min, int(img.shape[0] * w_min / img.shape[1])),
+                                    interpolation = interpolation)
+                        for img in img_list]
```
robertklee: I know this isn't your change, but the formatting for this is a bit hard to read.

robertklee:
I think `person_keypoints_val2017.json` was checked in accidentally @Airborne95

run_evaluation.py
```diff
@@ -90,4 +90,77 @@
...
+
+# h = HourglassNet(NUM_COCO_KEYPOINTS,DEFAULT_NUM_HG,INPUT_CHANNELS,INPUT_DIM,OUTPUT_DIM)
+representative_set_df = pd.read_pickle(os.path.join(DEFAULT_PICKLE_PATH, 'representative_set.pkl'))
+# train_df, val_df = h.load_and_filter_annotations(DEFAULT_TRAIN_ANNOT_PATH,DEFAULT_VAL_ANNOT_PATH,0.1)
+
+subdir = '2021-03-29-11h-07m_batchsize_16_hg_4_loss_weighted_mse_aug_light_sigma4_learningrate_5.0e-03_opt_rmsProp_gt-4kp_activ_linear_subset_0.50_wmse-5'
+eval = Evaluation(
+    # ensure model_json and weights files exist in current directory and num_hg_blocks matches model_json
+    model_sub_dir = subdir
+    ,epoch = 15)
```
robertklee:
```suggestion
    model_sub_dir = subdir,
    epoch = 15)
```

robertklee (commented):
Left a few comments. Looking forward to getting some quantitive insight into our models!

evaluation.py
```diff
@@ -146,6 +147,101 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+        Parameters
+        ----------
+        metadata : object
+        should be metadata associated to a single image
+
+        untransformed_predictions : list
+        a list of precitions that need to be transformed
+        Example:  [1,2,0,1,4,666,32...]
+    """
+    def undo_bounding_box_transformations(self, metadata, untransformed_predictions):
```
robertklee, line 187: just to confirm, have you validated this function against some ground truth data? (Take the transformed annotations, scatter plot on the transformed image. This is our baseline (what we feed to the model during train/val). Then take the same transformed annotations and run it through the `undo_bounding_box_transformations` and scatter plot on the original image. Maybe even plot the original annotations. They should line up).

evaluation.py
Airborne95: Good catch, I think this might have happened during a rebase!

requirements.txt
Airborne95: I tested it with both 1.17.0 and 1.20.0 and the numpy version has to be less than 1.20.0. I would suggest using 1.17.0 here.

evaluation.py
Airborne95: Nice one!

data_generator.py
Airborne95: Using that notation is returning a syntax error and I'm not entirely sure why. I have chosen to keep it as is for the time being.

evaluation.py
Airborne95: No that was not my intention, good catch!

run_evaluation.py
Airborne95: I think `run_evaluation.py` is just an example of how to use the methods introduced in this PR. I think it may be better suited to add a function that automates saving files with unique names in a second smaller PR given the time constraints. What do you think?

evaluation.py
Airborne95: Do you mean `heatmaps_to_keypoints` should return it as a 1d array?

evaluation.py
Airborne95: The if condition here is based on values of i and not values of `untransformed_predictions[i]` so I don't think we should face that issue as long as our input is always the same size. I do agree that it could be cleaned up in a future refactor, I am hesitant to refactor this now as it would require testing that I cannot do at the moment.

evaluation.py
Airborne95: I tried to clean it up :)

evaluation.py
Airborne95: Yes I did have a chance to test these changes. I used ground truth keypoints that had been transformed, then I called theses functions on them to make sure I could get back to the ground truth.

Airborne95:
Merge conflict based on an edit made about readability, when changes are approved I can fix it and merge!

evaluation.py
```diff
@@ -147,6 +149,98 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+            predict_heatmaps= self.predict_heatmaps(X_batch)
+            for X, metadata in zip(X_batch, metadatas):
+                keypoints = self.heatmaps_to_keypoints(predict_heatmaps[self.num_hg_blocks-1, j, :, :, :])
+                metadata = self.undo_bounding_box_transformations(metadata, keypoints)
+                list_of_predictions.append(self.create_oks_obj(metadata))
+                image_ids.append(metadata['src_set_image_id'])
+                j+=1
+        self.write_to_json_file(list_of_predictions, location)
+        return image_ids
+# ----------------------- End of Class -----------------------
```
nicolepeverley, line 241: Helpful 👌

nicolepeverley (approved):
Looks good! We may need to adjust how the json is saved for each evaluation run for OKS but we can follow up with that 👍

requirements.txt
robertklee: we may need to update the training script then. is it obvious if the numpy version is wrong? Like is it a crash, or does it fail silently and produce inaccurate results?

evaluation.py
robertklee: awesome, sounds good. :)

run_evaluation.py
robertklee: Yep, sounds like a good plan.

evaluation.py
```diff
@@ -147,6 +149,98 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
+        metadata['predicted_labels'] = predicted_labels
+        metadata['score'] = float(np.mean(np.array(list_of_scores)))
+        return metadata
+
+    def write_to_json_file(self, list_of_predictions, location):
+        f = open(location, "w")
+        f.write('[')
+        for i in range(len(list_of_predictions)):
+            f.write(json.dumps(list_of_predictions[i]))
+            if(i < len(list_of_predictions) - 1):
```
robertklee, line 213: nit:
```suggestion
            if i < len(list_of_predictions) - 1:
```

data_generator.py
robertklee: Oh it's because the above had a typo on the first line. my bad.
```
ann['src_set_image_id'
```
should be
```
ann['src_set_image_id'],
```

robertklee: I just tested and it no longer gives syntax error.
```suggestion
                metadata = {
                    "src_set_image_id":         ann['src_set_image_id'],
                    "ann_id":                   ann['ann_id'],
                    "coco_url":                 ann['coco_url'],
                    "cropped_width":            cropped_width,
                    "cropped_height":           cropped_height,
                    "anchor_x":                 anchor_x,
                    "anchor_y":                 anchor_y,
                    "input_dim":                self.input_dim,
                    "output_dim":               self.output_dim,
                    "transformed_label":        transformed_label, #DEBUG
                    "ground_truth_keypoints":   ann['keypoints'] #DEBUG
                }
```

robertklee (approved):
LGTM! Some small nits but they're not essential.

requirements.txt
Airborne95: Its a crash, doesn't run 😿

julianrocha (approved):
lgtm

requirements.txt
robertklee: Ok, that's good then. I was worried if we run in Colab (because they have pre-installed versions of all the software), it might silently fail. We can add more parameters to our training script to install a specific version of `numpy`if we encounter any issues.

data_generator.py
Airborne95: Good catch! I should have caught that too

Airborne95:
Thanks everyone for the reviews! Merging 🚀

PR #26, opened 2021-03-31
Auto Visualization

wanze-zhang:
auto_run.py is an extension of Julian's run_evaluation.py. It automatically runs the representative set for all models that are contained in a path and subdirectories within the path. 

Small changes in constants.py to work with os.path.join().

### DIFF
--- a/auto_run.py
+++ b/auto_run.py
@@ -1,0 +1,148 @@
+#%% 
+import hourglass
+import imp
+imp.reload(hourglass)
+from hourglass import HourglassNet
+from constants import *
+import matplotlib.pyplot as plt
+import os
+
+import data_generator
+imp.reload(data_generator)
+import data_generator
+import time
+
+import re
+import pandas as pd
+import evaluation
+from evaluation import Evaluation
+imp.reload(evaluation)
+from HeatMap import HeatMap
+
+#import HeatMap
+#imp.reload(HeatMap)
+
+import numpy as np
+
+#%% find duplicates in list
+
+def find_dup(pair_list, element):
+    for (a,b) in pair_list:
+        if element == a or element == b:
+            return True
+    return False
+
+#%% 
+
+def find_epochs(base_dir, sub_dir, epoch_dic, visited_models):
+    directory = os.listdir(os.path.join(base_dir, sub_dir))
+    weight_file = [f for f in directory if (f.endswith(".hdf5"))]
+    for name in weight_file:
+        match = re.search('hpe_epoch(\d+)', name)
+        if match:
+            epoch.append((match.group(1), sub_dir))
+
+    resume_file = [f for f in os.listdir(base_dir) if (sub_dir in f and f is not sub_dir)]
+    for file in resume_file:
+        visited_models.append(file)
+        for name in os.listdir(os.path.join(base_dir, file)):
+            match = re.search('hpe_epoch(\d+)', name)
+            #dup = find_dup(epoch, match.group(1))
+            #print("match is ", match, "dup is ", dup)
+            if match and not find_dup(epoch, match.group(1)):
+                epoch.append((match.group(1), file))
+
+    return list(set(epoch))
+
+
+#%% save stacked evaluation heatmaps
+
+def stacked_eval_heatmaps(test_df,eval):
+
+    generator = data_generator.DataGenerator(
+            df=test_df,
+            base_dir=DEFAULT_VAL_IMG_PATH,
+            input_dim=INPUT_DIM,
+            output_dim=OUTPUT_DIM,
+            num_hg_blocks=eval.num_hg_blocks,
+            shuffle=False,  
+            batch_size=len(test_df),
+            online_fetch=False)
+
+    # Select image to predict heatmaps
+    X_batch, y_stacked = generator[0] # There is only one batch in the generator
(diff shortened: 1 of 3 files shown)


auto_run.py
```diff
@@ -0,0 +1,148 @@
...
+
+#%% 
+
+def find_epochs(base_dir, sub_dir, epoch_dic, visited_models):
+    directory = os.listdir(os.path.join(base_dir, sub_dir))
+    weight_file = [f for f in directory if (f.endswith(".hdf5"))]
+    for name in weight_file:
+        match = re.search('hpe_epoch(\d+)', name)
+        if match:
+            epoch.append((match.group(1), sub_dir))
```
robertklee: Recommend dictionary here
```suggestion
    epochs = {}
    directory = os.listdir(os.path.join(base_dir, sub_dir))
    weight_file = [f for f in directory if (f.endswith(".hdf5"))]
    for name in weight_file:
        match = re.search('^hpe_epoch(\d+)', name)
        if match:
            epochs[match.group] = sub_dir
```

wanze-zhang: Oh yeah good suggestion :)

auto_run.py
```diff
@@ -0,0 +1,146 @@
...
+    X_batch, y_stacked = generator[168] # choose one image for evaluation
+    y_batch = y_stacked[0] # take first hourglass section
+    X, y = X_batch[0], y_batch[0] # take first example of batch
+
+    # Get predicted heatmaps for image
+    predict_heatmaps=eval.predict_heatmaps(X_batch)
+
+    # Get predicted keypoints from last hourglass (eval.num_hg_blocks-1)
+    keypoints = eval.heatmaps_to_keypoints(predict_heatmaps[eval.num_hg_blocks-1, 0, :, :, :])
+    print(keypoints)
```
nicolepeverley, line 103: Do we want to print the keypoints here?

nicolepeverley (commented):
Are we planning on using auto_run.py as a callback function for after each epoch or run only at the end for all epochs?

auto_run.py
```diff
@@ -0,0 +1,146 @@
...
+def stacked_eval_heatmaps(test_df,eval):
+
+    generator = data_generator.DataGenerator(
+            df=test_df,
+            base_dir=DEFAULT_VAL_IMG_PATH,
+            input_dim=INPUT_DIM,
+            output_dim=OUTPUT_DIM,
+            num_hg_blocks=eval.num_hg_blocks,
+            shuffle=False,  
+            batch_size=len(test_df),
```
julianrocha, line 67: If we only ever plan on passing the representative set, then this fine, but if we pass a `test_df` of the scale similar to our test set then it won't fit in memory. Will have to visualize in batches.

auto_run.py
```diff
@@ -0,0 +1,146 @@
...
+    plt.clf()
+    # Plot predicted keypoints on bounding box image
+    x = []
+    y = []
+    for i in range(NUM_COCO_KEYPOINTS):
+        if(keypoints[i,0] != 0 and keypoints[i,1] != 0):
+            x.append(keypoints[i,0])
+            y.append(keypoints[i,1])
+    plt.scatter(x,y)
+    plt.imshow(img)
```
julianrocha, line 119: If we are running this on all our models, then we may want to save these images to disk, we also may want to visualize the representative set rather than just `generator[168]`

auto_run.py
```diff
@@ -0,0 +1,146 @@
...
+representative_set_df = pd.read_pickle(os.path.join(DEFAULT_PICKLE_PATH, 'representative_set.pkl'))
+models = [] # keep track of the models that are visited 
+epoch = [] # extract out all epoch numbers (across different files) from model logs
+output_models = os.listdir(DEFAULT_OUTPUT_BASE_DIR) # keep track of the models have been explored before
+
+for sub_dir in os.listdir(DEFAULT_MODEL_BASE_DIR):
+    if '_hg_' in sub_dir and sub_dir not in models and sub_dir not in output_models:
+        models.append(sub_dir)
+        epoch = find_epochs(DEFAULT_MODEL_BASE_DIR, sub_dir, epoch, models)
+        for (n_epoch, model_file) in epoch:
```
julianrocha, line 136: Since the models make large progress at the beginning and less so near the end, and because loading/visualizing takes some time, we may want to limit the epochs we visualize. So for example, only visualize epochs: `[1,2,3,4,5,10,20,30,40]` or something along those lines

auto_run.py
wanze-zhang: I think we should just past in the representative set for quick comparisons. It takes about 40 mins to run 10 epochs on the representative set for each model.

auto_run.py
wanze-zhang: That is a very good point! I'll look into it :)

auto_run.py
wanze-zhang: Emmm, not necessary when we are running multiple models but it's very useful when we want to dig into one image. The key point np.array tells us which joint the model is seeing.

wanze-zhang: This is not a callback function. We are running this at the very end. Enabling a callback function would significantly increase our running time.

Ckoelewyn (approved):
LGTM

PR #27, opened 2021-03-31
Add ability to load any arbitrary image, centered on human, of any aspect ratio

robertklee:
This PR mainly adds support for loading arbitrary images and visualizing the predicted keypoints.

This PR refactors the bounding box transformation logic out of the data gen class to reuse for arbitrary img. I removed the arbitrary aspect ratio functionality because it was getting too complicated to get it working. 

## TODO
If anyone has capacity, please consider helping with these: 

* Verify that `ImageOps.exif_transpose(img)` is not needed in our data gen. If the PIL Image.read sees EXIF data about rotations applied to the image, it doesn't apply those, so we may be feeding rotated 90 degree images to the model. We should validate on the validation dataset that this EXIF is not present.
* It would be nice to pass in a directory of images, and the we auto read and predict on them
* visualizing heat maps doesn't work here yet. Not sure why. I tried to get to the bottom of this but ran out of time.

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -190,7 +190,7 @@ def load_and_preprocess_img(img_path, num_hg_blocks, bbox=None):
     X_batch = np.expand_dims(new_img.astype('float'), axis=0)
 
     # Add dummy heatmap "ground truth", duplicated 'num_hg_blocks' times
-    y_batch = [np.zeros((1, *(OUTPUT_DIM), NUM_COCO_KEYPOINTS)) for _ in range(num_hg_blocks)]
+    y_batch = [np.zeros((1, *(OUTPUT_DIM), NUM_COCO_KEYPOINTS), dtype='float') for _ in range(num_hg_blocks)]
 
     # Normalize input image
     X_batch /= 255
--- a/evaluation.py
+++ b/evaluation.py
@@ -1,16 +1,17 @@
 import os
 import re
+
 import cv2
-from scipy.ndimage import gaussian_filter, maximum_filter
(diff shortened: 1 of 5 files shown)


evaluation.py
```diff
@@ -158,29 +158,38 @@ def _non_max_supression(self, plain, threshold, windowSize=3):
...
-    if x is None or y is None or w is None or h is None:
-        # If any of the parameters are missing, crop a square area from top left of image
-        smaller_dim = img.size[0] if img.size[0] <= img.size[1] else img.size[1]
-        bbox = [0, 0, smaller_dim, smaller_dim]
-    else:
+    # Required because PIL will read EXIF tags about rotation by default. We want to
+    # preserve the input image rotation so we manually apply the rotation if required.
+    # See https://stackoverflow.com/questions/4228530/pil-thumbnail-is-rotating-my-image/
+    # and the answer I used: https://stackoverflow.com/a/63798032
+    img = ImageOps.exif_transpose(img)
```
nicolepeverley, line 168: This sounds good, have you tested to see if it changes the bbox in anyway? (Just thinking about the need to reverse)

robertklee: I haven't but I think we're good here because:
1. This is put in to ensure that the image we feed to the model is the right orientation (the one we see if we open it in a file viewer). I'd say it's safe to assume that the bounding box is created respective to the right orientation. 
2. We might not need this change in data generator which is what is used to train the model.

julianrocha (approved):
lgtm, square bbox is a reasonable constraint

PR #28, opened 2021-04-01
Add PCK Evaluation

nicolepeverley:
This PR adds a PCK evaluation metric for more insight into percentage of correct keypoints for each COCO keypoint label.

Example output from the representative set:

Percentage of Correct Key Points (PCK)

Nose:            0.80
Left Eye:        0.60
Right Eye:       0.60
Left Ear:        0.70
Right Ear:       0.80
Left Shoulder:   0.40
Right Shoulder:  0.40
Left Elbow:      0.60
Right Elbow:     0.50
Left Wrist:      0.40
Right Wrist:     0.60
Left Hip:        0.50
Right Hip:       0.50
Left Knee:       0.60
Right Knee:      0.70
Left Ankle:      0.40
Right Ankle:     0.60

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -242,8 +242,8 @@ class Evaluation():
     # This function evaluates PCK@0.2 == Distance between predicted and true joint < 0.2 * torso diameter
     # The PCK_THRESHOLD constant can be updated to adjust this threshold
     # https://github.com/cbsudux/Human-Pose-Estimation-101#percentage-of-correct-key-points---pck
-    def pck_eval(self, generator, location, annotation_file):
-        _, list_of_predictions = self.predict_keypoints(generator, location)
+    def pck_eval(self, generator, save_location, annotation_file):
+        _, list_of_predictions = self.predict_keypoints(generator, save_location)
         f = open(annotation_file)
         data = json.load(f)
 
@@ -294,63 +294,63 @@ class Evaluation():
             # Each image may have more than one annotation we need to check
             annotations = int(len(dist_list)/NUM_COCO_KEYPOINTS)
 
-            nose_correct = []
-            left_eye_correct = []
-            right_eye_correct = []
-            left_ear_correct = []
-            right_ear_correct = []
-            left_shoulder_correct = []
-            right_shoulder_correct = []
-            left_elbow_correct = []
-            right_elbow_correct = []
-            left_wrist_correct = []
-            right_wrist_correct = []
-            left_hip_correct = []
-            right_hip_correct = []
-            left_knee_correct = []
-            right_knee_correct = []
-            left_ankle_correct = []
-            right_ankle_correct = []
+            nose_correct = False
+            left_eye_correct = False
+            right_eye_correct = False
+            left_ear_correct = False
+            right_ear_correct = False
+            left_shoulder_correct = False
+            right_shoulder_correct = False
+            left_elbow_correct = False
+            right_elbow_correct = False
+            left_wrist_correct = False
+            right_wrist_correct = False
+            left_hip_correct = False
+            right_hip_correct = False
+            left_knee_correct = False
+            right_knee_correct = False
+            left_ankle_correct = False
+            right_ankle_correct = False
 
             # Append True to correct joint list if distance is below threshold for any annotation
             for j in range(annotations):
                 base = j*NUM_COCO_KEYPOINTS
-                nose_correct.append(dist_list[0+base]<=threshold)
-                left_eye_correct.append(dist_list[1+base]<=threshold)
-                right_eye_correct.append(dist_list[2+base]<=threshold)
-                left_ear_correct.append(dist_list[3+base]<=threshold)
-                right_ear_correct.append(dist_list[4+base]<=threshold)
-                left_shoulder_correct.append(dist_list[5+base]<=threshold)
-                right_shoulder_correct.append(dist_list[6+base]<=threshold)
-                left_elbow_correct.append(dist_list[7+base]<=threshold)
-                right_elbow_correct.append(dist_list[8+base]<=threshold)
-                left_wrist_correct.append(dist_list[9+base]<=threshold)
-                right_wrist_correct.append(dist_list[10+base]<=threshold)
-                left_hip_correct.append(dist_list[11+base]<=threshold)
-                right_hip_correct.append(dist_list[12+base]<=threshold)
-                left_knee_correct.append(dist_list[13+base]<=threshold)
-                right_knee_correct.append(dist_list[14+base]<=threshold)
-                left_ankle_correct.append(dist_list[15+base]<=threshold)
-                right_ankle_correct.append(dist_list[16+base]<=threshold)
+                if dist_list[0+base]<=threshold: nose_correct = True
+                if dist_list[1+base]<=threshold: left_eye_correct = True
+                if dist_list[2+base]<=threshold: right_eye_correct = True
+                if dist_list[3+base]<=threshold: left_ear_correct = True
+                if dist_list[4+base]<=threshold: right_ear_correct = True
+                if dist_list[5+base]<=threshold: left_shoulder_correct = True
+                if dist_list[6+base]<=threshold: right_shoulder_correct = True
+                if dist_list[7+base]<=threshold: left_elbow_correct = True
+                if dist_list[8+base]<=threshold: right_elbow_correct = True
+                if dist_list[9+base]<=threshold: left_wrist_correct = True
+                if dist_list[10+base]<=threshold: right_wrist_correct = True
+                if dist_list[11+base]<=threshold: left_hip_correct = True
+                if dist_list[12+base]<=threshold: right_hip_correct = True
+                if dist_list[13+base]<=threshold: left_knee_correct = True
+                if dist_list[14+base]<=threshold: right_knee_correct = True
+                if dist_list[15+base]<=threshold: left_ankle_correct = True
+                if dist_list[16+base]<=threshold: right_ankle_correct = True
 
             # Add one to correct keypoint count if any annotation was below threshold for image
-            if True in nose_correct: correct_keypoints["nose"] += 1
-            if True in left_eye_correct: correct_keypoints["left_eye"] += 1
-            if True in right_eye_correct: correct_keypoints["right_eye"] += 1
-            if True in left_ear_correct: correct_keypoints["left_ear"] += 1
-            if True in right_ear_correct: correct_keypoints["right_ear"] += 1
-            if True in left_shoulder_correct: correct_keypoints["left_shoulder"] += 1
-            if True in right_shoulder_correct: correct_keypoints["right_shoulder"] += 1
-            if True in left_elbow_correct: correct_keypoints["left_elbow"] += 1
-            if True in right_elbow_correct: correct_keypoints["right_elbow"] += 1
-            if True in left_wrist_correct: correct_keypoints["left_wrist"] += 1
-            if True in right_wrist_correct: correct_keypoints["right_wrist"] += 1
-            if True in left_hip_correct: correct_keypoints["left_hip"] += 1
-            if True in right_hip_correct: correct_keypoints["right_hip"] += 1
-            if True in left_knee_correct: correct_keypoints["left_knee"] += 1
-            if True in right_knee_correct: correct_keypoints["right_knee"] += 1
-            if True in left_ankle_correct: correct_keypoints["left_ankle"] += 1
-            if True in right_ankle_correct: correct_keypoints["right_ankle"] += 1
+            if nose_correct: correct_keypoints["nose"] += 1
+            if left_eye_correct: correct_keypoints["left_eye"] += 1
+            if right_eye_correct: correct_keypoints["right_eye"] += 1
+            if left_ear_correct: correct_keypoints["left_ear"] += 1
+            if right_ear_correct: correct_keypoints["right_ear"] += 1
+            if left_shoulder_correct: correct_keypoints["left_shoulder"] += 1
+            if right_shoulder_correct: correct_keypoints["right_shoulder"] += 1
+            if left_elbow_correct: correct_keypoints["left_elbow"] += 1
+            if right_elbow_correct: correct_keypoints["right_elbow"] += 1
+            if left_wrist_correct: correct_keypoints["left_wrist"] += 1
+            if right_wrist_correct: correct_keypoints["right_wrist"] += 1
+            if left_hip_correct: correct_keypoints["left_hip"] += 1
+            if right_hip_correct: correct_keypoints["right_hip"] += 1
+            if left_knee_correct: correct_keypoints["left_knee"] += 1
+            if right_knee_correct: correct_keypoints["right_knee"] += 1
+            if left_ankle_correct: correct_keypoints["left_ankle"] += 1
+            if right_ankle_correct: correct_keypoints["right_ankle"] += 1
             dist_list = []
 
         samples = len(list_of_predictions)
--- a/evaluation.py
+++ b/evaluation.py
@@ -239,11 +239,15 @@ class Evaluation():
         self.write_to_json_file(list_of_predictions, location)
         return image_ids, list_of_predictions
 
+    # This function evaluates PCK@0.2 == Distance between predicted and true joint < 0.2 * torso diameter
+    # The PCK_THRESHOLD constant can be updated to adjust this threshold
+    # https://github.com/cbsudux/Human-Pose-Estimation-101#percentage-of-correct-key-points---pck
     def pck_eval(self, generator, location, annotation_file):
         _, list_of_predictions = self.predict_keypoints(generator, location)
-        f = open(annotation_file,)
+        f = open(annotation_file)
         data = json.load(f)
 
+        # This function depends on the keypoints order listed in constants COCO_KEYPOINT_LABEL_ARR
         dist_list = []
         correct_keypoints = {
             "nose": 0,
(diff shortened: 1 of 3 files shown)


evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
...
+            for i, ann in enumerate(data['annotations']):
+                if data['annotations'][i]['image_id'] == prediction_image_id:
+                    annotation_keypoints = data['annotations'][i]['keypoints']
+                    prediction_keypoints = np.array(prediction_keypoints)
+                    annotation_keypoints = np.array(annotation_keypoints)
+
+                    # Joint at 11 is left hip, Joint at 12 is right hip
+                    left_hip_point = np.array(annotation_keypoints[33], annotation_keypoints[34])
+                    right_hip_point = np.array(annotation_keypoints[36], annotation_keypoints[37])
+                    # PCK@0.2 == Distance between predicted and true joint < 0.2 * torso diameter
```
nicolepeverley: I used this threshold for now because it was listed here https://github.com/cbsudux/Human-Pose-Estimation-101#percentage-of-correct-key-points---pck not sure if we want to use another threshold or what they mean by 50% of the head bone link haha

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
...
+                    correct_keypoints["left_hip"] += dist_list[11]<threshold
+                    correct_keypoints["right_hip"] += dist_list[12]<threshold
+                    correct_keypoints["left_knee"] += dist_list[13]<threshold
+                    correct_keypoints["right_knee"] += dist_list[14]<threshold
+                    correct_keypoints["left_ankle"] += dist_list[15]<threshold
+                    correct_keypoints["right_ankle"] += dist_list[16]<threshold
+                    dist_list = []
+
+        samples = len(list_of_predictions)
+        print("Percentage of Correct Key Points (PCK)\n")
```
nicolepeverley, line 357: I don't know if we want to list all of the joints here, but I thought it would be a good insight into the model. This code relies heavily on the order of the keypoints so this may have to be adjusted for augmentation.

robertklee: We won't be using augmentation on the validation set, so this won't be a problem :)

robertklee: But adding a note at the top of the function might be a good call

evaluation.py
robertklee: Sounds good to me! Maybe throw the threshold in `constants`?

nicolepeverley: So it adjusts the torso length for each image, I could add just the 0.2 to `constants`, what do you think?

robertklee: Sounds good, or even at the top of this file too. Just so it's a bit easier to find in the future

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
                 image_ids.append(metadata['src_set_image_id'])
                 j+=1
         self.write_to_json_file(list_of_predictions, location)
-        return image_ids
+        return image_ids, list_of_predictions
+
+    def pck_eval(self, generator, location, annotation_file):
+        _, list_of_predictions = self.predict_keypoints(generator, location)
+        f = open(annotation_file,)
```
robertklee: is the comma intentional?

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
...
+            "left_elbow": 0,
+            "right_elbow": 0,
+            "left_wrist": 0,
+            "right_wrist": 0,
+            "left_hip": 0,
+            "right_hip": 0,
+            "left_knee": 0,
+            "right_knee": 0,
+            "left_ankle": 0,
+            "right_ankle": 0
```
robertklee, line 269: would bools make more sense here? `False`?

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
...
+        for prediction in list_of_predictions:
+            prediction_image_id = prediction['image_id']
+            prediction_keypoints = prediction['keypoints']
+            for i, ann in enumerate(data['annotations']):
+                if data['annotations'][i]['image_id'] == prediction_image_id:
+                    annotation_keypoints = data['annotations'][i]['keypoints']
+                    prediction_keypoints = np.array(prediction_keypoints)
+                    annotation_keypoints = np.array(annotation_keypoints)
+
+                    # Joint at 11 is left hip, Joint at 12 is right hip
```
robertklee: It took me a minute to figure out the indexing, so adding this suggestion here :)

```suggestion
                    # Joint at 11 is left hip, Joint at 12 is right hip. Multiply by 3 as each keypoint has (x, y, confidence/visibility) to get the array index
```

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
...
+                    # PCK@0.2 == Distance between predicted and true joint < 0.2 * torso diameter
+                    torso = np.linalg.norm(left_hip_point-right_hip_point)
+                    threshold = 0.2*torso
+
+                    for i in range(50):
+                        if i % 3 == 0:
+                            prediction_point = np.array(prediction_keypoints[i], prediction_keypoints[i+1])
+                            annotation_point = np.array(annotation_keypoints[i], annotation_keypoints[i+1])
+                            dist = (np.linalg.norm(prediction_point-annotation_point))
+                            dist_list.append(dist)
```
robertklee: Could we do: 
```suggestion
                    for i in range(NUM_COCO_KEYPOINTS):
                        base = i * 3
                        prediction_point = np.array(prediction_keypoints[base], prediction_keypoints[base+1])
                        annotation_point = np.array(annotation_keypoints[base], annotation_keypoints[base+1])
                        dist = (np.linalg.norm(prediction_point-annotation_point))
                        dist_list.append(dist)
```

evaluation.py
robertklee: Never mind, I understand how this is used now. This is across the entire set, right?

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
                 image_ids.append(metadata['src_set_image_id'])
                 j+=1
         self.write_to_json_file(list_of_predictions, location)
-        return image_ids
+        return image_ids, list_of_predictions
```
julianrocha, line 240: This additional return value may affect the OKS script, in `run_evaluation.py` line 147

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
...
+                    correct_keypoints["left_elbow"] += dist_list[7]<threshold
+                    correct_keypoints["right_elbow"] += dist_list[8]<threshold
+                    correct_keypoints["left_wrist"] += dist_list[9]<threshold
+                    correct_keypoints["right_wrist"] += dist_list[10]<threshold
+                    correct_keypoints["left_hip"] += dist_list[11]<threshold
+                    correct_keypoints["right_hip"] += dist_list[12]<threshold
+                    correct_keypoints["left_knee"] += dist_list[13]<threshold
+                    correct_keypoints["right_knee"] += dist_list[14]<threshold
+                    correct_keypoints["left_ankle"] += dist_list[15]<threshold
+                    correct_keypoints["right_ankle"] += dist_list[16]<threshold
```
robertklee: For PCK, do we count the model predicting **no joint** as success? (I feel like you've answered this before but I can't remember)

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
...
+        print("Left Elbow: {:.2f}".format(correct_keypoints["left_elbow"]/samples))
+        print("Right Elbow: {:.2f}".format(correct_keypoints["right_elbow"]/samples))
+        print("Left Wrist: {:.2f}".format(correct_keypoints["left_wrist"]/samples))
+        print("Right Wrist: {:.2f}".format(correct_keypoints["right_wrist"]/samples))
+        print("Left Hip: {:.2f}".format(correct_keypoints["left_hip"]/samples))
+        print("Right Hip: {:.2f}".format(correct_keypoints["right_hip"]/samples))
+        print("Left Knee: {:.2f}".format(correct_keypoints["left_knee"]/samples))
+        print("Right Knee: {:.2f}".format(correct_keypoints["right_knee"]/samples))
+        print("Left Ankle: {:.2f}".format(correct_keypoints["left_ankle"]/samples))
+        print("Right Ankle: {:.2f}".format(correct_keypoints["right_ankle"]/samples))
```
robertklee: small nit: using a tab or spaces might make this row easier to read once it prints out

evaluation.py
```diff
@@ -237,7 +237,97 @@ def predict_keypoints(self,generator, location):
                 image_ids.append(metadata['src_set_image_id'])
                 j+=1
         self.write_to_json_file(list_of_predictions, location)
-        return image_ids
+        return image_ids, list_of_predictions
+
+    def pck_eval(self, generator, location, annotation_file):
```
robertklee: is `location` a save location or read location?

robertklee (approved):
LGTM! Some comments, mostly on adding a few lines of comments here and there.

julianrocha (approved):
The code looks great, I agree with Roberts style comments. I left a comment about an addition that may affect OKS. I'm not super versed with the math behind PCK so it's tricky for me to give feedback on the implementation. Overall this is great, didn't think we would fit in PCK so ty for pushing this through. Gonna be great for our prezi/report and general tuning insight. I like that we can see a breakdown by kp so we can see which ones we are struggling with. P.S. Not sure if you did this but might be worth testing this on ground truth data to ensure it gives 100% PCK

evaluation.py
nicolepeverley: Yeah this is keeping track of how many times each keypoint is "correct" and then it divides by the number of predictions were looking at!

evaluation.py
nicolepeverley: ahh forgot to add that :)

evaluation.py
nicolepeverley: nope whoops good catch!

evaluation.py
nicolepeverley: I believe it should only take the first return value, but I will double check!

evaluation.py
nicolepeverley: So, yes it will see how far away it was from the true point. So if our model predicts 0, 0 and the ground truth is 0, 0 it will definitely be under the threshold and count as a correct point.

evaluation.py
nicolepeverley: It's used to save the location of the predicted keypoints json in `predict_keypoints`, I used the function Rafay had already made to return the predicted keypoints as well as saving them to the json for OKS.

evaluation.py
nicolepeverley: Good catch I will fix that 👍

evaluation.py
robertklee: That sounds good! Is the absence of a joint encoded as 0,0 for both ground truth and predicted?

evaluation.py
robertklee: Small nit: could you add `save` or something similar to indicate it's for writing?

evaluation.py
```diff
@@ -237,7 +237,143 @@ def predict_keypoints(self,generator, location):
...
+
+                    for i in range(NUM_COCO_KEYPOINTS):
+                        base = i*3
+                        prediction_point = np.array(prediction_keypoints[base], prediction_keypoints[base+1])
+                        annotation_point = np.array(annotation_keypoints[base], annotation_keypoints[base+1])
+                        dist = (np.linalg.norm(prediction_point-annotation_point))
+                        dist_list.append(dist)
+
+            # Each image may have more than one annotation we need to check
+            annotations = int(len(dist_list)/NUM_COCO_KEYPOINTS)
```
nicolepeverley, line 295: Okay so I ran into a bit of an issue! When I did some more testing I realized that there may be more than one annotation per image. So I've changed it so it will check if it's within the threshold for any of the annotations it finds. This can probably be handled more gracefully but let me know what you guys think 👍

evaluation.py
nicolepeverley: Yep it should be.

evaluation.py
robertklee: Sounds like a good v1! Maybe add a note somewhere (to backlog, maybe?) so we don't forget.

evaluation.py
```diff
@@ -237,7 +237,143 @@ def predict_keypoints(self,generator, location):
...
+                left_elbow_correct.append(dist_list[7+base]<=threshold)
+                right_elbow_correct.append(dist_list[8+base]<=threshold)
+                left_wrist_correct.append(dist_list[9+base]<=threshold)
+                right_wrist_correct.append(dist_list[10+base]<=threshold)
+                left_hip_correct.append(dist_list[11+base]<=threshold)
+                right_hip_correct.append(dist_list[12+base]<=threshold)
+                left_knee_correct.append(dist_list[13+base]<=threshold)
+                right_knee_correct.append(dist_list[14+base]<=threshold)
+                left_ankle_correct.append(dist_list[15+base]<=threshold)
+                right_ankle_correct.append(dist_list[16+base]<=threshold)
```
robertklee: I think this could be made cleaner if you set them as `false` bools in the beginning and then in the for loop you run a `joint_x = joint_x or (joint <= threshold)`

evaluation.py
```diff
@@ -237,7 +237,143 @@ def predict_keypoints(self,generator, location):
...
+            if True in left_elbow_correct: correct_keypoints["left_elbow"] += 1
+            if True in right_elbow_correct: correct_keypoints["right_elbow"] += 1
+            if True in left_wrist_correct: correct_keypoints["left_wrist"] += 1
+            if True in right_wrist_correct: correct_keypoints["right_wrist"] += 1
+            if True in left_hip_correct: correct_keypoints["left_hip"] += 1
+            if True in right_hip_correct: correct_keypoints["right_hip"] += 1
+            if True in left_knee_correct: correct_keypoints["left_knee"] += 1
+            if True in right_knee_correct: correct_keypoints["right_knee"] += 1
+            if True in left_ankle_correct: correct_keypoints["left_ankle"] += 1
+            if True in right_ankle_correct: correct_keypoints["right_ankle"] += 1
```
robertklee: And then here we can just say `if right_knee_correct: correct_keypoints[right_knee] += 1`

evaluation.py
nicolepeverley: Oh yeah you know what it would probably make more sense if theyre just left as bools

robertklee: There might even be some nice `numpy` way to do this on the entire vector at once haha but it's all good if we don't

Ckoelewyn (approved):
LGTM

evaluation.py
```diff
@@ -237,7 +237,143 @@ def predict_keypoints(self,generator, location):
...
+
+                    # Calculate PCK@0.2 threshold for image
+                    # Joint at 11 is left hip, Joint at 12 is right hip. Multiply by 3 as each keypoint has (x, y, visibility) to get the array index
+                    left_hip_point = np.array(annotation_keypoints[33], annotation_keypoints[34])
+                    right_hip_point = np.array(annotation_keypoints[36], annotation_keypoints[37])
+                    torso = np.linalg.norm(left_hip_point-right_hip_point)
+                    threshold = PCK_THRESHOLD*torso
+
+                    for i in range(NUM_COCO_KEYPOINTS):
+                        base = i*3
```
robertklee: just realized we have a `const` for this
```suggestion
                        base = i * NUM_COCO_KP_ATTRBS
```

evaluation.py
nicolepeverley: Yeah haha I'm sure there's a better way to do it, it just gets finicky with different numbers of annotations for each image

evaluation.py
```diff
@@ -237,7 +237,143 @@ def predict_keypoints(self,generator, location):
...
+                if dist_list[7+base]<=threshold: left_elbow_correct = True
+                if dist_list[8+base]<=threshold: right_elbow_correct = True
+                if dist_list[9+base]<=threshold: left_wrist_correct = True
+                if dist_list[10+base]<=threshold: right_wrist_correct = True
+                if dist_list[11+base]<=threshold: left_hip_correct = True
+                if dist_list[12+base]<=threshold: right_hip_correct = True
+                if dist_list[13+base]<=threshold: left_knee_correct = True
+                if dist_list[14+base]<=threshold: right_knee_correct = True
+                if dist_list[15+base]<=threshold: left_ankle_correct = True
+                if dist_list[16+base]<=threshold: right_ankle_correct = True
```
robertklee: this would do the same thing (and is cleaner), but I'm good either way!


```suggestion
                nose_correct            = nose_correct              or dist_list[0+base]  <= threshold
                left_eye_correct        = left_eye_correct          or dist_list[1+base]  <= threshold
                right_eye_correct       = right_eye_correct         or dist_list[2+base]  <= threshold
                left_ear_correct        = left_ear_correct          or dist_list[3+base]  <= threshold
                right_ear_correct       = right_ear_correct         or dist_list[4+base]  <= threshold
                left_shoulder_correct   = left_shoulder_correct     or dist_list[5+base]  <= threshold
                right_shoulder_correct  = right_shoulder_correct    or dist_list[6+base]  <= threshold
                left_elbow_correct      = left_elbow_correct        or dist_list[7+base]  <= threshold
                right_elbow_correct     = right_elbow_correct       or dist_list[8+base]  <= threshold
                left_wrist_correct      = left_wrist_correct        or dist_list[9+base]  <= threshold
                right_wrist_correct     = right_wrist_correct       or dist_list[10+base] <= threshold
                left_hip_correct        = left_hip_correct          or dist_list[11+base] <= threshold
                right_hip_correct       = right_hip_correct         or dist_list[12+base] <= threshold
                left_knee_correct       = left_knee_correct         or dist_list[13+base] <= threshold
                right_knee_correct      = right_knee_correct        or dist_list[14+base] <= threshold
                left_ankle_correct      = left_ankle_correct        or dist_list[15+base] <= threshold
                right_ankle_correct     = right_ankle_correct       or dist_list[16+base] <= threshold
```

PR #29, opened 2021-04-01
Fix R/L bug in augmentation pipeline

julianrocha:
When we perform R/L augmentation, a right wrist becomes a left wrist but we do not account for this. Here is an example of the bug before these changes:
With the R/L flip (clearly wrong)
image
Without the R/L flip
image

This PR extends the aug pipeline to do custom R/L flip, and reshuffles the kp array accordingly.

### DIFF
--- a/data_augmentation.py
+++ b/data_augmentation.py
@@ -3,6 +3,7 @@
 from util import validate_enum
 from imgaug.augmenters.meta import OneOf
 import numpy as np # linear algebra
+import random
 
 # import the library and helpers
 import imageio
@@ -11,6 +12,7 @@ from imgaug import augmenters as iaa
 from imgaug.augmentables.kps import Keypoint, KeypointsOnImage
 
 from constants import ImageAugmentationStrength
+from constants import RL_FLIP
 
 # Holy resources: https://nbviewer.jupyter.org/github/aleju/imgaug-doc/blob/master/notebooks/B01%20-%20Augment%20Keypoints.ipynb
 # Credit to the above notebook for their tutorial on keypoint augmentation
@@ -42,7 +44,6 @@ def get_augmenter_pipeline(strength_enum):
     #### Options:
     ImageAugmentationStrength.heavy:
         - Blur with probability 30% with sigma              2.0
-        - Horizontal flip (left to right) with probability  50%
         - Up to 3 of the following:
             - Sharpening                                    0.8 to 1.2, blending factor between (0,1)
             - Hue and saturation change of                  -30 to 30 out of 255
@@ -57,7 +58,6 @@ def get_augmenter_pipeline(strength_enum):
 
     ImageAugmentationStrength.medium:
         - Blur with probability 30% with sigma              1.5
-        - Horizontal flip (left to right) with probability  50%
         - Up to 3 of the following:
             - Sharpening                                    0.85 to 1.15, blending factor between (0,0.5)
             - Hue and saturation change of                  -20 to 20 out of 255
@@ -72,7 +72,6 @@ def get_augmenter_pipeline(strength_enum):
 
     ImageAugmentationStrength.light:
         - Blur with probability 30% with sigma              1
-        - Horizontal flip (left to right) with probability  50%
         - Up to 2 of the following:
             - Sharpening                                    None
             - Hue and saturation change of                  -15 to 15 out of 255
@@ -202,8 +201,6 @@ def get_augmenter_pipeline(strength_enum):
     aug_pipeline = iaa.Sequential([
         # apply Gaussian blur with a sigma between 0 and x to 30% of the images
         iaa.Sometimes(0.3, iaa.GaussianBlur((0, iaaGaussianBlurSigmaMax))),
-        # horizontally flip 50% of the time
-        iaa.Sometimes(0.5, iaa.Fliplr(1.0)),
         iaa.SomeOf((0, iaaMaxNumberImageAppearanceOperations),
             iaaSomeOfImageAppearance
         ),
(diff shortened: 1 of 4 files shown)


robertklee:
Have you noticed L/R labels being related to the actual person's left/right, or whether it's left/right from our view?

julianrocha:
> Have you noticed L/R labels being related to the actual person's left/right, or whether it's left/right from our view?

It's the person's left/right. The ground truth image example with no flip in this PR demonstrates that.

data_augmentation.py
```diff
@@ -223,6 +220,15 @@ def get_augmenter_pipeline(strength_enum):
 
     return aug_pipeline
 
+def flipRL(image, keypoints):
+    if random.random() < RL_FLIP:
+        return image, keypoints
+    flip = iaa.Fliplr(1.0)
+    flipped_img, flipped_kps = flip(image=image, keypoints=keypoints)
+    flat_map = lambda f, xs: [y for ys in xs for y in f(ys)] # https://dev.to/turbaszek/flat-map-in-python-3g98
```
robertklee, line 227: Love this!!

data_generator_tests.py
```diff
@@ -79,16 +85,16 @@ def display_img(annId):
...
-heatmap = y[:,:,1]
-print(X.max(), X.min(), X.mean())
-print(y.max(), y.min(), y.mean())
-hm = HeatMap(X,heatmap)
-hm.plot(transparency=0.5,show_axis=True,show_colorbar=True)
+for i in range(7,8):
+  print(COCO_KEYPOINT_LABEL_ARR[i])
+  heatmap = y[:,:,i]
+  hm = HeatMap(X,heatmap)
+  hm.plot(transparency=0.5,show_axis=True,show_colorbar=True)
```
robertklee: Small nit on indenting

robertklee (commented):
I'm out right now but I will take a closer look in an hour or so. So far looks great! I love the lambda.

robertklee:
> > Have you noticed L/R labels being related to the actual person's left/right, or whether it's left/right from our view?
> 
> 
> 
> It's the person's left/right. The ground truth image example with no flip in this PR demonstrates that.

That was what confused me, because it seems like the top image should be the person's left wrist, right?

Edit: never mind, it looks right now

data_augmentation.py
```diff
@@ -223,6 +220,15 @@ def get_augmenter_pipeline(strength_enum):
...
+    flipped_fixed_kps = [flipped_kps[0]] + flat_map(lambda pair : (pair[1],pair[0]), zip(flipped_kps[1::2],flipped_kps[2::2]))
+    # Leave nose unchanged, flip every pair of eye, ear, etc...
+    return flipped_img, flipped_fixed_kps
```
nicolepeverley, line 230: This looks great to me good catch! Just to confirm you're flipping it so that it remains in the keypoints order in constants?

nicolepeverley (approved):

data_augmentation.py
julianrocha: Yes, so:
`["nose","left_eye","right_eye","left_ear","right_ear","left_shoulder","right_shoulder","left_elbow","right_elbow","left_wrist","right_wrist","left_hip","right_hip","left_knee","right_knee","left_ankle","right_ankle"]`
becomes:
`["nose","right_eye","left_eye","right_ear","left_ear"...]`

This is necessary b/c when you mirror flip an image, a person's right side becomes their left side, so after fixing the coords, we need to swap labels with the other side.

nicolepeverley: Ahh gotcha I wonder how this would work if we feed an augmented image into OKS or PCK if we can't directly compare to the ground truth.

data_augmentation.py
```diff
@@ -11,6 +12,7 @@
 from imgaug.augmentables.kps import Keypoint, KeypointsOnImage
 
 from constants import ImageAugmentationStrength
+from constants import RL_FLIP
```
robertklee: Nit,merge the two lines

data_augmentation.py
robertklee: We aren't applying augmentation to our validation or test set (and i'm not sure if we should). If we do, we can hold on to the augmented ground truth keypoints and use that as our ground truth. We will have to avoid running PCK on the training generator without accounting for this though

data_augmentation.py
robertklee: tip: you can right click in VSCode and there's actually a "sort imports" button that merges it for you

data_generator.py
```diff
@@ -268,6 +268,9 @@ def __getitem__(self, idx):
                 # Perform data augmentation randomly
                 image_aug, kpsoi_aug = self.augmenter(image=transformed_img, keypoints=kpsoi)
 
+                # Perform a R/L augmentation randomly
```
robertklee: This is a bit wordy, so feel free to reword.

```suggestion
                # Perform a R/L augmentation randomly, applying R/L flip to the labels as well to maintain the right order 
```

data_generator_tests.py
```diff
@@ -56,19 +56,25 @@ def display_img(annId):
...
 import data_generator
 imp.reload(data_generator)
 from data_generator import DataGenerator
 
+
+representative_set_df = pd.read_pickle(os.path.join(DEFAULT_PICKLE_PATH, 'representative_set.pkl'))
 generator = DataGenerator(
-    df=val_df,
+    df=representative_set_df,
     base_dir=DEFAULT_TRAIN_IMG_PATH,
```
robertklee: Does `TRAIN_IMG_PATH` work here? (I think they're the same as `VAL_IMG_PATH`, right?)

data_generator_tests.py
```diff
@@ -79,16 +85,16 @@ def display_img(annId):
...
-heatmap = y[:,:,1]
-print(X.max(), X.min(), X.mean())
-print(y.max(), y.min(), y.mean())
-hm = HeatMap(X,heatmap)
-hm.plot(transparency=0.5,show_axis=True,show_colorbar=True)
+for i in range(NUM_COCO_KEYPOINTS):
+  print(COCO_KEYPOINT_LABEL_ARR[i])
+  heatmap = y[:,:,i]
+  hm = HeatMap(X,heatmap)
+  hm.plot(transparency=0.5,show_axis=True,show_colorbar=True)
```
robertklee, line 99: small nit on indenting

robertklee (approved):
Looks good, and this is a super quick turnaround on this bug! 🚀

data_augmentation.py
robertklee: We aren't applying augmentation to our validation or test set (and i'm not sure if we should). If we do, we can hold on to the augmented ground truth keypoints and use that as our ground truth. We will have to avoid running PCK on the training generator without accounting for this though

data_augmentation.py
julianrocha: Didn't know this existed ty!

data_generator_tests.py
julianrocha: They're the same but Imma change to  `VAL_IMG_PATH ` b/c it's more clear

robertklee:
🚀 turnaround time! Looks like it's time to restart my training runs haha

PR #30, opened 2021-04-03
Add script to clean empty training runs. Some small general code cleanup

robertklee:
This PR mainly adds a script that cleans trivial training runs from the `models` and `logs` directories. The script will only **move** files rather than **delete** them. It moves these files to a subdirectory named `trivial-runs`.

We can then manually verify that the folders contain no files. For tensor board, I recommend downloading the `trivial-runs` folder and running `tensorboard --logdir logs/trivial-runs` and checking that there is no data.

This PR also updates some minor code styling and fixes a few small bugs.


clean_training_artifacts.py
```diff
@@ -0,0 +1,78 @@
...
+def move_trivial_runs(base_dir, non_trivial_runs, verbose=True):
+    # Create temporary folder
+    DESTINATION_SUBDIR = 'trivial-runs'
+
+    temp_dir = os.path.join(base_dir, DESTINATION_SUBDIR)
+    if not os.path.exists(temp_dir):
+        os.makedirs(temp_dir)
+
+    print(f'Moving empty training runs in "{base_dir}" to "{temp_dir}". Proceed?')
+    input("Press Enter to continue...")
```
Ckoelewyn, line 51: No option for not proceeding, do we need this?

robertklee: This is mostly there to confirm it has the directory names correct (in case there's a hidden `/` at the front or something). If they don't want to continue, they can exit the program with `CTRL C`. That was my original idea, at least

robertklee:
I have run this on my personal Drive and validated that this cleans a lot of empty training Tensorboard logs and empty subdir properly. Runs that were started in the first 1-2 days of training being possible (before March 13th) are not found properly because the logs and models dir had different names at that point, but I don't think this should cause any issues for anyone because those runs were purely experimental.

robertklee:
In any case, I'd still double check the folders before you delete anything. You can download the entire `logs/trivial-runs` folder and run it in tensor board. There shouldn't be any data that is plotted.

julianrocha (approved):
lgtm!

PR #34, opened 2021-04-08
Re-enable functionality to run on any user specified image

robertklee:
Updated: This PR re-enables functionality to run on any specified image or file directory. It also adds the ability to average predictions from normal and horizontally flipped images to achieve higher accuracy.

TODO:
Surface image as variable instead of saving it as file


julianrocha (approved):
nice!

PR #35, opened 2021-04-11
Fix bug where finding the max heatmap location after non-maximum suppression returned empty list

robertklee:
`np.where(peaks == peaks.max())` was returning an empty list for some reason. I suspect it is due to floating point equality comparison.

This PR uses a more robust way that guarantees that a result will be returned.

I've also looked into partially patching `PCK`.

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -324,6 +324,8 @@ class Evaluation():
 
                     # TODO figure out what to do if a hip isn't present
                     threshold = DEFAULT_PCK_THRESHOLD
+
+                    # If both hips are present
                     if annotation_keypoints[35] > 0 and annotation_keypoints[38] > 0:
                         left_hip_point = np.array(annotation_keypoints[33], annotation_keypoints[34])
                         right_hip_point = np.array(annotation_keypoints[36], annotation_keypoints[37])
--- a/evaluation.py
+++ b/evaluation.py
@@ -323,10 +323,15 @@ class Evaluation():
 
                     # Calculate PCK@0.2 threshold for image
                     # Joint at 11 is left hip, Joint at 12 is right hip. Multiply by 3 as each keypoint has (x, y, visibility) to get the array index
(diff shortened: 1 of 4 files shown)


nicolepeverley (approved):
LGTM 👍  the only other change I had was using the head as a secondary measure but we could also just roll with this for now

robertklee:
@nicolepeverley i like your changes better. Let's merge both and keep yours for PCK-related fixes for now? I know Corey was also taking a look at it from the original paper.

evaluation.py
```diff
@@ -229,18 +229,16 @@ def heatmaps_to_keypoints(self, heatmaps, threshold=HM_TO_KP_THRESHOLD):
...
             # Choose the max point in heatmap (we only pick 1 keypoint in each heatmap)
             # and get its coordinates and confidence
-            y, x = np.where(peaks == peaks.max())
+            y, x = np.unravel_index(np.argmax(peaks), peaks.shape)
 
-            if int(x[0]) > 0 and int(y[0]) > 0:
-                x_new = int(x[0])
-                y_new = int(y[0])
-                conf_new = peaks[y_new, x_new]
+            if peaks[y, x] > HM_TO_KP_THRESHOLD / 5:
```
julianrocha: Why divide this by 5? Can we just change the constant?

julianrocha (approved):
Nice! So for the PCK stuff we will roll with Nicoles PR? I left one comment but looks good otherwise

evaluation.py
robertklee: I wasn't sure if non-maximum suppression would reduce the max intensity so i picked an arbitrary number. anything below that threshold should have already been reset to 0 though

robertklee:
yeah, the pck still is not fixed with this PR. See attached.

[pck.csv](https://github.com/robertklee/COCO-Human-Pose/files/6292501/pck.csv)

PR #36, opened 2021-04-11
Fix PCK when no torso exists

nicolepeverley:
Added conditions for when a torso doesn't exist in an annotation and when a head doesn't exist in an annotation for PCK.

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -316,14 +316,15 @@ class Evaluation():
         for prediction in list_of_predictions:
             prediction_image_id = prediction['image_id']
             prediction_keypoints = prediction['keypoints']
-            for i, ann in enumerate(data['annotations']):
+            anns = len(data['annotations'])
+            for i in range(anns):
                 if data['annotations'][i]['image_id'] == prediction_image_id:
                     annotation_keypoints = data['annotations'][i]['keypoints']
                     prediction_keypoints = np.array(prediction_keypoints)
                     annotation_keypoints = np.array(annotation_keypoints)
 
                     # Calculate PCK@0.2 threshold for image
-                    # TODO figure out what to do if a hip isn't present
+                    # TODO figure out what to do if a hip or ear isn't present
                     threshold = DEFAULT_PCK_THRESHOLD
 
                     # If both hips are present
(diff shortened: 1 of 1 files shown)


evaluation.py
```diff
@@ -315,18 +315,30 @@ def pck_eval(self, list_of_predictions):
...
                     prediction_keypoints = np.array(prediction_keypoints)
                     annotation_keypoints = np.array(annotation_keypoints)
 
                     # Calculate PCK@0.2 threshold for image
                     # Joint at 11 is left hip, Joint at 12 is right hip. Multiply by 3 as each keypoint has (x, y, visibility) to get the array index
-                    left_hip_point = np.array(annotation_keypoints[33], annotation_keypoints[34])
-                    right_hip_point = np.array(annotation_keypoints[36], annotation_keypoints[37])
-                    torso = np.linalg.norm(left_hip_point-right_hip_point)
-                    threshold = PCK_THRESHOLD*torso
+                    if((annotation_keypoints[33] != 0 and annotation_keypoints[34] != 0) or (annotation_keypoints[36] != 0 and annotation_keypoints[37] != 0)):
```
robertklee: I think checking the visibility flag is more robust (and shorter) 
```suggestion
                    if annotation_keypoints[35] > 0 and annotation_keypoints[38] > 0:
```

evaluation.py
```diff
@@ -315,18 +315,30 @@ def pck_eval(self, list_of_predictions):
...
+                        left_hip_point = np.array(annotation_keypoints[33], annotation_keypoints[34])
+                        right_hip_point = np.array(annotation_keypoints[36], annotation_keypoints[37])
+                        torso = np.linalg.norm(left_hip_point-right_hip_point)
+                        threshold = PCK_THRESHOLD*torso
+                    # Use head threshold if no torso exists
+                    # Joint at 1 is left ear, Joint at 2 is right ear. Multiply by 3 as each keypoint has (x, y, visibility) to get the array index
+                    elif((annotation_keypoints[3] != 0 and annotation_keypoints[4] != 0) or (annotation_keypoints[6] != 0 and annotation_keypoints[7] != 0)):
```
robertklee: same here
```suggestion
                    elif annotation_keypoints[5] > 0 and annotation_keypoints[8] > 0:
```

robertklee (approved):
left a small comment. lgtm otherwise

evaluation.py
nicolepeverley: oh good call I forgot about the flag!

evaluation.py
```diff
@@ -315,18 +315,32 @@ def pck_eval(self, list_of_predictions):
...
+                        torso = np.linalg.norm(left_hip_point-right_hip_point)
+                        threshold = PCK_THRESHOLD*torso
+                    # Use head threshold if no torso exists
+                    # Joint at 1 is left ear, Joint at 2 is right ear. Multiply by 3 as each keypoint has (x, y, visibility) to get the array index
+                    # Check visibility flags for both ear joints
+                    elif(annotation_keypoints[5] > 0 and annotation_keypoints[8] > 0):
+                        left_ear_point = np.array(annotation_keypoints[3], annotation_keypoints[4])
+                        right_ear_point = np.array(annotation_keypoints[6], annotation_keypoints[7])
+                        head = np.linalg.norm(left_ear_point-right_ear_point)
+                        threshold = PCK_THRESHOLD*head
```
julianrocha, line 345: Since head len is always less than torso len, should we multiply by a small factor (`ex: 1.5`) to make it a closer estimate of the torso. I'm not sure what other PCK implementations have done.

julianrocha (approved):
Nice! left one comment

evaluation.py
nicolepeverley: This is a good question because I was thinking it would usually use the head threshold when it's closer up to the person   if there isn't a torso in the image. I haven't been able to find too much documentation on it, but I'm good either way.

nicolepeverley: I'll add these changes for now and we can always update it if we find more docs!

robertklee: sounds good!

PR #37, opened 2021-04-12
Fix PCK evaluation bug by using annotation ID instead of image ID

robertklee:
So far, this fix has produced correct values on the representative set, with epoch 1 having 0.26, epoch 5 as 0.64, and epoch 107 as 0.74.
PCK_graph_hg4_no_flip

### DIFF
--- a/evaluation.py
+++ b/evaluation.py
@@ -289,11 +289,11 @@ class Evaluation():
     # The PCK_THRESHOLD constant can be updated to adjust this threshold
     # https://github.com/cbsudux/Human-Pose-Estimation-101#percentage-of-correct-key-points---pck
     def pck_eval(self, list_of_predictions):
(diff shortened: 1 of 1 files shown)


evaluation.py
```diff
@@ -577,6 +544,7 @@ def _undo_bounding_box_transformations(self, metadata, untransformed_predictions
     def _create_oks_obj(self, metadata):
         oks_obj = {}
         oks_obj["image_id"] = int(metadata['src_set_image_id'])
+        oks_obj["ann_id"] = int(metadata['ann_id'])
```
nicolepeverley, line 547: I didn't realize the original image was associated with an ann_id nice catch!

nicolepeverley (approved):
LGTM 💯  This makes way more sense if we have access to the ann_id right away nice bug squashing 🐛

PR #39, opened 2021-04-21
Update with successful deployment config, skeleton colouring, image display, docs

robertklee:
This PR mainly fixes our deployment. Full change log below:

- the ability to deploy our app by adding necessary requirements and other config files
- changes the skeleton linkage colour to be constant between different predictions, whereas before, the colour would be dynamically selected. This makes it easier to compare which particular joint link was detected
- removes the axes from our predicted keypoint images
- updates the `README` for better visibility. For deployment, the `numpy` requirement had to be relaxed in `requirements.txt`. I've added a troubleshooting line in the README to fix the `numpy` versioning issue
- adds max file size requirement of 20 MB, which puts us just under the `PIL` limitation of ~10,000x10,000 pixels (`Image.MAX_IMAGE_PIXELS`) (see [this Stack Overflow question](https://stackoverflow.com/questions/56174099/how-to-load-images-larger-than-max-image-pixels-with-pil))

### DIFF
--- a/.streamlit/config.toml
+++ b/.streamlit/config.toml
@@ -1,3 +1,3 @@
 [server]
-maxUploadSize=50
+maxUploadSize=20
 headless=true
--- a/.streamlit/config.toml
+++ b/.streamlit/config.toml
@@ -1,2 +1,3 @@
 [server]
 maxUploadSize=20
+headless = true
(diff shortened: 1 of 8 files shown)


wanze-zhang (approved):
This is great! Thank you for the deployment ☺️👏

